PR #35, opened 2021-01-28
Remind PR merge and new case failure 

KivenChen:
Added features:
- Comment PRs with failed test case with "issue closed" tag
- Comment PRs with failed test case that has not been recorded in the database

Refinement:
- Data types, ci report query fix, ...

This PR is the current up and running version started on Jan. 27th.

PR #37, opened 2021-02-02
Feature/new case comments

AsterNighT:
Comment to PRs which encounter brand new failure cases when doing scheduling case fetch.


AsterNighT:
Closed due to duplication with #35

PR #39, opened 2021-02-04
Add github actions for pr and push to main

AsterNighT:

PR #40, opened 2021-02-04
Update test path in parser

AsterNighT:
Pls check this before #39 
## Why
The tests in **parser** package now use path relative to /sync_ci, which make it have to be compiled and run manually from /sync_ci, rather than be processed along with other tests (although there's none for now) by `go test ./...`.
## What does this PR do
Change path to be relative to the /sync_ci/pkg/parser folder, namely to the test file it self.


lilinghai:
pls resolve the conflict

lilinghai (approved):
LGTM

PR #44, opened 2021-02-25
Fetch jenkins pipelines

AsterNighT:
Signed-off-by: AsterNighT <klxjt99@outlook.com>


AsterNighT:
wait a minute, pd_test is.fetched twice so there is an issue

PR #46, opened 2021-02-26
docker_rust: install gcc 8

BusyJay:
We need gcc 8 to build latest grpc.

### DIFF
--- a/jenkins/Dockerfile/centos7_golang-1.13_rust
+++ b/jenkins/Dockerfile/centos7_golang-1.13_rust
@@ -5,14 +5,15 @@ MAINTAINER Liu Yin <liuy@pingcap.com>
 USER root
 WORKDIR /root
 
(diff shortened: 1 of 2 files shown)


jenkins/Dockerfile/centos7_golang-1.13_rust
```diff
@@ -5,14 +5,15 @@ MAINTAINER Liu Yin <liuy@pingcap.com>
...
-	&& cd cmake-3.10.0 \
-	&& ./bootstrap \
-	&& make \
-	&& make install \
+RUN yum install --nogpgcheck -y centos-release-scl \
+	&& yum install -y devtoolset-8 \
+	&& yum clean all
+
+RUN yum remove -y cmake \
+	&& curl -L https://github.com/Kitware/CMake/releases/download/v3.10.0/cmake-3.10.0-linux-x86_64.sh -o cmake-3.10.0.sh \
```
BusyJay, line 13: I can't download cmake from the given IP, so I change it to the official link.

PR #51, opened 2021-02-27
*: use source directly to enable devtoolset-8

BusyJay:
We can't use scl_enable before
https://github.com/sclorg/scl-utils/pull/35 is released.


zhouqiang-cl (approved):
LGTM

PR #155, opened 2021-04-28
make tidb_title_check more strict

disksing:
Disallow terrible titles like `foo:do something` or `: do something`


zhouqiang-cl (approved):
LGTM

PR #156, opened 2021-04-28
Fix bug when build tidb & plugins, tidb src not exist

purelind:
Fix a bug: when build tidb and tidb plugins, remove deleteDir when start build plugins.

### DIFF
--- a/jenkins/pipelines/cd/optimization-build-tidb-linux-amd.groovy
+++ b/jenkins/pipelines/cd/optimization-build-tidb-linux-amd.groovy
@@ -69,10 +69,11 @@ try {
                 }
             }
             builds["Build tidb && plugins"] = {
+                def ws = pwd()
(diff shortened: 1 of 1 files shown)


jenkins/pipelines/cd/optimization-build-tidb-linux-amd.groovy
```diff
@@ -115,8 +116,7 @@ try {
                         }
                     }
                     container("golang") {
```
zhouqiang-cl, line 118: Do we need to remove container("golang")

purelind: Nodes can be defined in a pipeline and then used, however, default execution always goes to the jnlp container. Specify the container we want to execute  task in is necessary and reasonable.

PR #180, opened 2021-05-17
Patch2 darwin build parallel

purelind:
Correct login error, if build cache exist then skip build. 
Build tiflash on mac with i5.


dianqihanwangzi:
LGTM

PR #181, opened 2021-05-17
fix bug: logic error, skip build if build cache existed

purelind:
Correct logic error, skip build if build cache existed


zhouqiang-cl (approved):
LGTM

PR #198, opened 2021-05-20
change git version from v1.8 to v2.22

CadmusJiang:
Signed-off-by: Cadmus <CadmusJiang@gmail.com>

### DIFF
--- a/jenkins/Dockerfile/centos7_golang-1.16
+++ b/jenkins/Dockerfile/centos7_golang-1.16
@@ -3,7 +3,9 @@ FROM hub.pingcap.net/jenkins/centos7_jenkins
 USER root
 WORKDIR /root
 
-RUN sudo yum install git -y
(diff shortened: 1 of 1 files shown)


jenkins/Dockerfile/centos7_golang-1.16
```diff
@@ -3,7 +3,9 @@ FROM hub.pingcap.net/jenkins/centos7_jenkins
 USER root
 WORKDIR /root
 
-RUN sudo yum install git -y
+RUN sudo yum remove git -y
+RUN sudo curl https://setup.ius.io | sh
+RUN sudo yum install -y git222
```
purelind, line 8: 1. Is sudo necessary?
2. Maybe we could use one RUN contains these cmd.

purelind (approved):
LGTM

PR #210, opened 2021-05-24
Put go1.16 logic code after release test params setting

purelind:
Put go1.16 logic code after release test params setting in case of job failed when triggered by qa_release.


zhouqiang-cl (approved):
LGTM

PR #211, opened 2021-05-24
fix some miss go1.16 node setting on some test step

purelind:
fix error when some ci test case not use go1.16, set test-slave node


zhouqiang-cl (approved):
LGTM

PR #214, opened 2021-05-25
BR: add upstream(tikv/tidb/pd) support

Leavrth:
select the specific test script for upstream

### DIFF
--- a/jenkins/pipelines/ci/br/br_ghpr_unit_and_integration_test.groovy
+++ b/jenkins/pipelines/ci/br/br_ghpr_unit_and_integration_test.groovy
@@ -416,6 +416,12 @@ catchError {
                                 "br_full",
                                 "br_gcs",
                                 "br_s3",
+                                "lightning_alter_random",
+                                "lightning_new_collation",
+                                "lightning_row-format-v2",
+                                "lightning_s3",
+                                "lightning_sqlmode",
+                                "lightning_tiflash",
                             ]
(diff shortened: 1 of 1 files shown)


zhouqiang-cl:
Did  this patch will reduce CI time?

Leavrth:
> Did this patch will reduce CI time?

PR Solve the second problem in the follow 2 problems:
  1. tikv/tidb/pd triggers `br_ghpr_unit_and_integration_test` asynchronously, which means it would be return immediately no matter whether `br_ghpr_unit_and_integration_test` is successed or not.
  2. `br_ghpr_unit_and_integration_test` will do the whole test for upstream tikv/tidb/pd which is unnecessary and waste  ci time.

We select specific test cases when tikv/tidb/pd triggers `br_ghpr_unit_and_integration_test` as upstream, which reduce CI time.

zhouqiang-cl:
> > Did this patch will reduce CI time?
> 
> PR Solve the second problem in the follow 2 problems:
> 
> 1. tikv/tidb/pd triggers `br_ghpr_unit_and_integration_test` asynchronously, which means it would be return immediately no matter whether `br_ghpr_unit_and_integration_test` is successed or not.
> 2. `br_ghpr_unit_and_integration_test` will do the whole test for upstream tikv/tidb/pd which is unnecessary and waste  ci time.
> 
> We select specific test cases when tikv/tidb/pd triggers `br_ghpr_unit_and_integration_test` as upstream, which reduce CI time.

got it, thank you

PR #223, opened 2021-05-27
upgrade tikv ci go version from 1.13 to 1.16.4, only for branch master

purelind:
-  tikv ci upgrade go version from 1.13 to 1.16.4, only for branch master
-  tikv test use agent pod with memory volume


zhouqiang-cl (approved):
LGTM

PR #224, opened 2021-05-27
add dm ci pipeline, upgrade go version

purelind:
- add dm ci pipeine
- dm ci upgrade go version from 1.13 to 1.16.4, branch master only


dianqihanwangzi:
lgtm

PR #232, opened 2021-05-28
ci: remove GO111MODULE=off in ticdc pipelines

amyangfei:
to fix error `go: modules disabled by GO111MODULE=off; see 'go help modules'`
gomodule is enabled by default in ticdc


purelind:
LGTM

PR #247, opened 2021-06-02
mv unit-test and leak-test to nightly-test

XuHuaiyu:
We move `unit test` and `leak test` to [nightly test](https://github.com/PingCAP-QE/ci/blob/main/jenkins/pipelines/nightly/tidb/tidb_nightly_unit_test.groovy) because of:
1. Leak test rarely failed.
2. Unit test is almost the same as what `check_dev_2` does. (Unit test sets `config.checkBeforeDropLDFlag=1`.)


zhouqiang-cl (approved):
LGTM

PR #257, opened 2021-06-09
[on tidb_unit_test] grep fail info of previous failed test

MartinNose:


jenkins/pipelines/ci/tidb/tidb_ghpr_unit_test.groovy
```diff
@@ -261,12 +261,18 @@ try {
...
                                 }
                             }catch (err) {
+                                println 
+                                sh """
+                                    cat test.log | cut -d" " -f2- | grep -Ev "^\\[[[:digit:]]{4}(/[[:digit:]]{2}){2}" | grep -A 30 "\\-------" | grep -A 29 "^FAIL:"
+                                """
+
                                 throw err
                             }finally {
+                                rm test.log
```
zhouqiang-cl, line 275: why rm

MartinNose: Because the log is already sent to stdout and just is teed to the `test.log` for the err handling to grep.

PR #273, opened 2021-06-15
disable Integration Connection Test on tidb/tikv/pd

purelind:
disable Integration Connection Test on tidb/tikv/pd


zhouqiang-cl (approved):
LGTM

PR #278, opened 2021-06-16
Update tidb_ghpr_check_2.groovy

disksing:
`store/tikv` is going to be removed. We need to update CI script.


zhouqiang-cl (commented):
LGTM

PR #282, opened 2021-06-16
tidb build trigger by upstream ci

purelind:
pr merged commit trigger tidb build ci, binary is required by other integration test


dianqihanwangzi:
lgtm

PR #297, opened 2021-06-23
Add TiDB-Operator related Dockerfile and Jenkins pipeline.

n0vad3v:
This PR will:
* Add TiDB-Operator related Dockerfile and Jenkins pipeline.
* tidb-backup-manager-arm64 Dockerfile is switched to `centos:8` to avoid libc related problems.

Before Merge, we should:
1. Change `baseUrl` in `jenkins/pipelines/cd/build-operator-arm-image.groovy` to use the right branch(instead of `operator-arm64`).
2. Make sure those images are working.


DanielZhangQD (approved):
LGTM

PR #327, opened 2021-07-09
BR: use fileserver instead stash, merge shell scripts to speed up, split little slow cases

Leavrth:
1. use fileserver instead stash. The compress step cannot be omitted so fileserver only speed up file transfer.
2. merge shell scripts. Each `sh` keyword in groovy will trigger master send `scripts` to slaver, which is a heavy workload. So merge these `download shell scripts` into one `sh` keyword in groovy and parallel downloading will reduce much CI time.
3. split little slow cases. Now it has much parallel nodes to do tests and some nodes finish tests much more quickly. In the best case, all nodes finish tests at the same time. So the PR try to reduce the granularity of way to split tests in order to be  more closer to the best case and reduce the number of nodes besides.


purelind (approved):
LGTM

lance6716:
cc @Ehco1996

PR #339, opened 2021-07-23
tidb: use memory volume in tidb_ghpr_integration_common_test

purelind:
* use memory volume in tidb integration test, releate issue [tidb #26458](https://github.com/pingcap/tidb/issues/26458)


zhouqiang-cl (approved):
LGTM

PR #357, opened 2021-08-05
tidb: build on pod with memory volume

purelind:
* build tidb on pod with memory volume, sometimes start tidb-server failed on the hdd disk


zhouqiang-cl (approved):
LGTM

PR #364, opened 2021-08-10
fix: use and operator to get release-xxx branch name

purelind:
* use && operator in case of panic on example branch not release-xxx


dianqihanwangzi:
lgtm

PR #411, opened 2021-09-07
tikv: hotfix branch use correct image

purelind:
* remove useless job data report api
* hotfix branch test use correct image, example release-4.0-20210724 use image hub.pingcap.net/jenkins/tikv-cached-release-4.0:latest


dianqihanwangzi (approved):
lgtm

PR #419, opened 2021-09-16
ticdc: use case name in integration step

amyangfei:
Currently integration test stage name is `integration test step_12`, we can't find the running case directly (have to click the stage name and expand details).

PR #421, opened 2021-09-17
Revert "ticdc: use case name in integration step"

amyangfei:
Reverts PingCAP-QE/ci#419

Coverage step uses hard coded step name, ref:
https://github.com/PingCAP-QE/ci/blob/09643537687a5d956bb3543095595fe51fa739bf/jenkins/pipelines/ci/ticdc/integration_test_common.groovy#L286-L289


hi-rustin (approved):
It will break coverage collection, very sorry 🧎, I wasn't able to test and take it into account.

PR #436, opened 2021-10-09
tidb: remove parser replace check for migration

xhebox:
Signed-off-by: xhe <xw897002528@gmail.com>

For #28644, which is blocked by this test.


tisonkun:
@xhebox I think it refers to https://github.com/pingcap/tidb/pull/28644?

PR #471, opened 2021-11-12
dockerfile: add debug dockerfile for qa test

purelind:
* add debug dockerfile for qa test, use centos7.9.2009 as base image


dianqihanwangzi (approved):
lgtm

PR #487, opened 2021-12-02
tikv ci feature branch use default image

purelind:
* tikv ci in feature branch to user default cached ci image


zhouqiang-cl:
LGTM

PR #592, opened 2022-01-11
cache tiflash dockerfile

SchrodingerZhu:
Signed-off-by: Schrodinger ZHU Yifan <i@zhuyi.fan>

We may still want to adjust some parameters in this dockerfile


zhouqiang-cl (approved):
LGTM

PR #608, opened 2022-01-13
tikv: support coverage

BusyJay:


BusyJay:
Tested and verify the coverage can be uploaded correctly. As it's supposed to run one time per day, I don't do anything to speed up the job. It takes about 1 hour to finish.

purelind:
> Tested and verify the coverage can be uploaded correctly. As it's supposed to run one time per day, I don't do anything to speed up the job. It takes about 1 hour to finish.

We can observe for a period of time to see if you can accelerate it. Maybe cargo cache or not enough CPU or memory resource.

jenkins/pipelines/ci/tikv/tikv_ghpr_coverage.groovy
```diff
@@ -0,0 +1,117 @@
...
+                sh label: 'Post-build: Generating coverage', script: """
+                cd \$HOME/tikv-src
+                grcov . --binary-path ./target/debug/ -s . -t lcov --branch --ignore-not-existing --ignore "/*" --ignore "target/debug/*" -o lcov.info
+                """
+
+                withCredentials([string(credentialsId: 'TIKV_CODECOV_TOKEN', variable: 'CODECOV_TOKEN')]) {
+                    timeout(180) {
+                        sh label: 'Post-build: Uploading coverage', script: """
+                        cd \$HOME/tikv-src
+                        curl -OLs https://uploader.codecov.io/latest/linux/codecov
```
purelind, line 87: codecov binary could be download from fileserver:
`curl -LO ${FILE_SERVER_URL}/download/cicd/ci-tools/codecov`

purelind (approved):

PR #611, opened 2022-01-13
compat w/ jenkins docker

SchrodingerZhu:
Docker with older versions does not support default args syntax iwth whitespace.
Signed-off-by: Schrodinger ZHU Yifan <i@zhuyi.fan>


zhouqiang-cl (approved):
LGTM

PR #754, opened 2022-02-15
use glusterfs as go mod cache(tidb ut)

dianqihanwangzi:
- use glusterfs as go mod cache(tidb ut)


dianqihanwangzi:
test： https://ci.pingcap.net/blue/organizations/jenkins/tidb_ghpr_unit_test/detail/tidb_ghpr_unit_test/35610/pipeline/

wuhuizuo:
current no needed, because we have refactored the pipeline

PR #767, opened 2022-02-22
DM CI use a pinned version of sync-diff

lance6716:
replay (running): https://ci2.pingcap.net/job/dm_ghpr_integration_test/3692/

currently the CI uses nightly version of sync-diff and it has some problem. In order to cope with the urgent release, I hardcoded a version of sync-diff. We will revert it later


lance6716:
close due to tidb-tools always use master after release-5.3

dianqihanwangzi (approved):
lgtm

PR #773, opened 2022-03-01
Cache TiFlash sanitizer script

SchrodingerZhu:
see https://ci.pingcap.net/blue/organizations/jenkins/tiflash-sanitizer-daily/activity


purelind (approved):
LGTM

PR #778, opened 2022-03-02
remove tiflash download url

zhouqiang-cl:
Since we have tiflash download on tiup, it seems no longer need to separate url to download tiflash 
Signed-off-by: Qiang Zhou <zhouqiang.cl@gmail.com>


wuhuizuo:
conflict and the changed file was deleted, close it.

PR #783, opened 2022-03-02
use master sync-diff since it has fixed bug

lance6716:
waiting CI replay

the fix is https://github.com/pingcap/tidb-tools/pull/597


lance6716:
@zhouqiang-cl @purelind 

PTAL

PR #799, opened 2022-03-10
[DNM] workaround for tikv image issue

purelind:
* tikv image issue: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory


dianqihanwangzi (approved):
Lgtm

PR #807, opened 2022-03-14
pd master ci use go1.18

purelind:
* master branch ci upgrade go version from go1.16 to go1.18


purelind:
releated to https://github.com/tikv/pd/issues/4710

PR #808, opened 2022-03-14
binlog ci use go1.18

purelind:
* master branch ci upgrade go version from go1.16 to go1.18

related to https://github.com/pingcap/tidb-binlog/pull/1142


purelind:
releated to https://github.com/pingcap/tidb-binlog/pull/1142

PR #814, opened 2022-03-17
tiflow ci use go1.18

purelind:
branch master upgrade go version from go1.16 to go1.18


purelind:
releated to https://github.com/pingcap/tiflow/pull/4910

PR #829, opened 2022-03-21
fix sync diff version

dianqihanwangzi:
- fix sync diff version


dianqihanwangzi:
replayed pipeline: https://ci2.pingcap.net/blue/organizations/jenkins/cdc_ghpr_integration_test/detail/cdc_ghpr_integration_test/10739/pipeline/132

purelind (approved):

PR #844, opened 2022-03-23
add install nc

better0332:


n0vad3v (changes requested):
Should not install debug tools in release Dockerfile.

better0332:
but amd64 environment have nc, it should be consistent

PR #940, opened 2022-05-07
ci: skip ticdc integration tests if matching engine dir

amyangfei:
ref: task-5 in https://github.com/pingcap/tiflow/issues/5352

Skip ticdc integration test/kafka integration test if only changes engine code.


amyangfei:
PTAL, thanks. @purelind

purelind (approved):
lgtm

PR #1007, opened 2022-05-23
[WIP] use new cluster

purelind:


wuhuizuo:
should use draft instead of naming the PR with `WIP` prefix in GitHub.

purelind:
> should use draft instead of naming the PR with `WIP` prefix in GitHub.

I tend to think of this as a specification of your past work, not a specification common to the open source community. Just style, not rule restrictions

Here are a number of pr for kubernetes: https://github.com/kubernetes/kubernetes/pulls?q=is%3Apr+%22WIP%22

PR #1017, opened 2022-05-25
jenkins/pipeline/cd: make curl exit with err code

wuhuizuo:
curl default always exit with zero code.
add `--fail` option to `curl` cmd for file uploading.

### DIFF
--- a/jenkins/pipelines/cd/multibranch/build_binlog_multi_branch.groovy
+++ b/jenkins/pipelines/cd/multibranch/build_binlog_multi_branch.groovy
@@ -126,10 +126,10 @@ try {
                     timeout(10) {
                         sh """
                         echo "${githash}" > sha1
-                        curl --fail -F ${refspath}=@sha1 ${FILE_SERVER_URL}/upload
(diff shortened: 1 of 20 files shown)


wuhuizuo:
@purelind pls review it

purelind:
--fail : Fail silently (no output at all) on HTTP errors (H)
Have you misunderstood its meaning?

purelind (changes requested):
--fail: Fail silently (no output at all) on HTTP errors (H)
not means  exit with an error code.

wuhuizuo:
> --fail: Fail silently (no output at all) on HTTP errors (H) not means exit with an error code.

pls see: 
image

wuhuizuo:
new pipeline test passed (by  manually replay) @purelind

wuhuizuo:
The `fileserver` will be replaced with a more standard artifact server in this year.

the `fileserver` is not RESTFul and cloud native, and I can not fix and rollout it online, current the best way is update the `cd` pipelines.

PR #1049, opened 2022-05-30
Remove the deprecated target page_ctl

hehechen:
After https://github.com/pingcap/tiflash/pull/5003, the target `page_ctl` is deprecated, need to remove it from CI.


JaySon-Huang (approved):
LGTM

purelind (approved):

PR #1057, opened 2022-05-31
feat(jenkins/pipelines): enhance tarball download with curl retry and continue options

wuhuizuo:
added options:
- `-C -`: to make it continue where it dropped of (if you run the command again). The dash afterwards tells it to figure out where to resume from.
- `--retry 3`: to retry 3 times.


wuhuizuo:
@purelind @ heibaijian please review it.

PR #1122, opened 2022-06-21
add some real tikv tests

purelind:
* currently those tests run on [github action](https://github.com/pingcap/tidb/blob/master/.github/workflows/integration-test-with-real-tikv.yml) . github action free quota is limited, tasks often encounter queues


purelind:
replay test ci 
1. https://ci.pingcap.net/blue/organizations/jenkins/tidb_ghpr_check_2/detail/tidb_ghpr_check_2/83186/pipeline/69/

PR #1131, opened 2022-06-27
Disable tikv-client unit test in TiSpark integration_test

Daemonxiao:
**What**: Disable tikv-client unit test in TiSpark integration_test.

**Why**: Because tikv-client unit test has been already done in GitHub Action, we need not add it to the integration test again.


purelind (approved):
LGTM

PR #1135, opened 2022-06-29
*: remove 2 mysql test files

zimulala:
Remove 2 mysql test files: `opt_hints_index` and `desc_index_innodb`

There are many tests in `opt_hints_index`  that are not supported by TIDB at present

The result is different for storage = tikv and unistore in `desc_index_innodb`


purelind (approved):
LGTM

PR #1136, opened 2022-06-30
mysql_test: using backlist in mysql_test.go instead of whitelist in CI

zimulala:
Using backlist in mysql_test.go instead of whitelist in CI.


purelind (approved):
LGTM

PR #1163, opened 2022-07-13
tikv: mysql-test use blacklist

purelind:


purelind:
replay test ci: https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_integration_common_test/detail/tikv_ghpr_integration_common_test/4049/pipeline/144/

PR #1231, opened 2022-08-04
split DM test case to speed up

lance6716:


lance6716:
https://ci2.pingcap.net/job/dm_ghpr_integration_test/8658/

purelind (approved):
LGTM

PR #1263, opened 2022-08-15
feat(build): upload tiup multi os-arch package parallel

purelind:
* upload different os-arch tiup packages in parallel


lijie0123 (approved):

purelind:
pipeline upload cost time decrease from 18 min to 8min

PR #1297, opened 2022-08-24
fix: curl cli retry and continue args

wuhuizuo:
- `--retry` retry
- `-C` continue
- `-f` or `--fail` fail with non-zero exit code


purelind (approved):
LGTM

PR #1322, opened 2022-08-31
pd master use go1.19

purelind:


purelind:
releated issue https://github.com/tikv/pd/issues/5454

PR #1323, opened 2022-08-31
feat(tiflow): master use go1.19

purelind:


purelind:
releated to https://github.com/pingcap/tiflow/issues/6909

PR #1324, opened 2022-08-31
feat(tidb-binlog): master use go1.19

purelind:


purelind:
releated to https://github.com/pingcap/tidb-binlog/issues/1203

PR #1334, opened 2022-09-02
feat: dm it use go1.19

purelind:


purelind:
hold unti https://github.com/pingcap/tiflow/pull/6961 merged

PR #1346, opened 2022-09-06
feat: download tidb-test from git instead of fileserver

purelind:
* get tidb-test code directly from git instead of fileserver


purelind:
replay ci https://ci.pingcap.net/blue/organizations/jenkins/tidb_ghpr_mysql_test/detail/tidb_ghpr_mysql_test/25162/pipeline/

PR #1359, opened 2022-09-13
Feat integration boost redo

Icemap:


purelind:
The current configuration of the tidb-test pipeline has been moved to the directory https://github.com/PingCAP-QE/ci/tree/main/pipelines/pingcap/tidb-test/latest. The old pipeline will no longer be maintained, so this PR should be closed.

PR #1394, opened 2022-09-21
tidb(ci): adjust keep days to 30, remove keep nums

purelind:
* currently tidb ci generates 1500 build histories per week, we need to keep history for 1 month.


wuhuizuo (approved):
/lgtm

PR #1412, opened 2022-09-23
tiflash: Only apply lint and format on master branch

JaySon-Huang:
As we upgrade the compiler version, the static lints checks and format could be changed after the compile toolchain is upgraded.
Only apply lints and format check on the master branch.

Changing `TiFlashTestBasic.cpp` in master don't raise any lint errors, but the cherry-pick to release-5.4 raise errors.
https://github.com/pingcap/tiflash/pull/5924


purelind (approved):
LGTM

PR #1457, opened 2022-10-10
upgrade cache image 20221010

hawkingrei:
Signed-off-by: Weizhen Wang <wangweizhen@pingcap.com>

PR #1459, opened 2022-10-10
feat(jobs/pingcap/tidb/latest): update bazel data mount src dir

wuhuizuo:


wuhuizuo:
it should be merge after #1457

lijie0123 (approved):

PR #1463, opened 2022-10-12
fix: not trigger build for currently task always failed

purelind:
not trigger ghpr_build on any branch , currently task always failed.


wuhuizuo (changes requested):
please delete the `ghpr_build` job and files

PR #1466, opened 2022-10-13
tiflash: improve error notice and fix idents

breezewish:
1. When build failed, will show a link pointing the build URL inline. Sometimes BlueOcean does not show the Trigger Build, and this could be helpful.
2. Fixed idents for some scripts. (If you want to view diff, please diff without whitespace)
3. Disable coverage in the unit test phase, as currently it does not work well. We can enable it later when the efficiency can be improved by running it in the post PR stage and providing more detailed coverages.
4. Adjusted unit test CPU request from 10c to 12c, since our test concurrency is set to 12. Also changed the limit from 20c to 12c, because 12c should be enough. (this limit change may be reverted later, if we really incur some problems)

### DIFF
--- a/jenkins/pipelines/ci/tiflash/tiflash-build-common.groovy
+++ b/jenkins/pipelines/ci/tiflash/tiflash-build-common.groovy
@@ -264,18 +264,17 @@ def checkoutStage(repo_path, checkout_target) {
         dir(repo_path) {
             def cache_path = "/home/jenkins/agent/ci-cached-code-daily/src-tics.tar.gz"
             if (fileExists(cache_path)) {
-                println "get code from nfs to reduce clone time"
-                sh """
-                set +x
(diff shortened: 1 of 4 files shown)


purelind:
please try to replay an existing ci pipeline to make sure those changes do not break ci

purelind (approved):
lgtm

purelind:
> please try to replay an existing ci pipeline to make sure those changes do not break ci

update
build-common https://ci.pingcap.net/blue/organizations/jenkins/tiflash-build-common/detail/tiflash-build-common/14179/pipeline

unit-test https://ci.pingcap.net/blue/organizations/jenkins/tiflash-ghpr-unit-tests/detail/tiflash-ghpr-unit-tests/6512/pipeline

wuhuizuo:
Good job!  Thanks for reducing resource requires and pipeline step optimization.

PR #1469, opened 2022-10-14
use cargo sort in tikv format

tabokie:
Signed-off-by: tabokie <xy.tao@outlook.com>

### DIFF
--- a/jenkins/pipelines/ci/tikv/tikv_ghpr_test.groovy
+++ b/jenkins/pipelines/ci/tikv/tikv_ghpr_test.groovy
@@ -180,7 +180,6 @@ stage("Prepare") {
                             cd \$HOME/tikv-src
                             export RUSTFLAGS=-Dwarnings
                             make format
-                            cargo sort || true
                             git diff --quiet || (git diff; echo Please make format and run tests before creating a PR; exit 1)
                         """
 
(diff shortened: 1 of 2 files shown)


jenkins/pipelines/ci/tikv/tikv_ghpr_test.groovy
```diff
@@ -179,7 +179,9 @@ stage("Prepare") {
                         sh label: 'Run lint: format', script: """
                             cd \$HOME/tikv-src
                             export RUSTFLAGS=-Dwarnings
-                            make format && git diff --quiet || (git diff; echo Please make format and run tests before creating a PR; exit 1)
+                            make format
```
Connor1996, line 182: why doesn't make format run cargo sort?

tabokie: Oh, I didn't notice it's already in the Makefile. Maybe it's because the CI environment doesn't have cargo sort.

jenkins/pipelines/ci/tikv/tikv_ghpr_test.groovy
```diff
@@ -179,7 +179,9 @@ stage("Prepare") {
...
+                            cargo sort || true
```
tabokie, line 183:
```suggestion
```

purelind (approved):
lgtm

jenkins/pipelines/ci/tikv/tikv_ghpr_test.groovy
Connor1996: pre-format installs cargo-sort automatically, why install it explicitly?

PR #1473, opened 2022-10-14
tiflash: fix trigger tiflash-build-common error

purelind:
* fix trigger downstream pipeline with error 
```shell
Cannot invoke method getResult() on null object
```


JaySon-Huang (approved):
LGTM

PR #1475, opened 2022-10-17
ci(tiflow): update make engine script

amyangfei:
ref: https://github.com/pingcap/tiflow/pull/7377

Tiflow will simplify binaries that are built from `make engine`.

Replace `make engine` with `make tiflow tiflow-demo`


CharlesCheung96 (approved):

wuhuizuo (approved):
LGTM

PR #1476, opened 2022-10-17
fix(libraries/tipipeline/vars): add support for feature branch PRs in component checkout

wuhuizuo:
- default trunk branch set to `master`, but caller can give custom value, as:
```groovy
component.checkout('xxx.git', 'depA', 'feature/abc', '', "", "custom-trunk-branch")
```


purelind (approved):
lgtm

PR #1484, opened 2022-10-18
feat(jobs): using `BUILD_URL` to set git status

wuhuizuo:
why:
- build url is faster than run display url.
- identified problems list in build page, some error can not display in blue ocean.


purelind:
Some error logs are not easy to locate quickly on build_url

wuhuizuo:
no need nore.

PR #1508, opened 2022-10-27
fix(pipelines/pingcap/tidb): bash regex for feature branch checkout third repos

wuhuizuo:


wuhuizuo:
replaying build: https://do.pingcap.net/jenkins/job/pingcap/job/tidb/job/ghpr_check2/1969/

PR #1542, opened 2022-11-11
update base image to rockylinux:9.0

lichunzhu:
as title, update tools base image to `rockylinux:9.0` for security.


jenkins/Dockerfile/release/debug-image/ticdc
```diff
@@ -1,4 +1,4 @@
-FROM pingcap/alpine-glibc:alpine-3.14.6
+FROM rockylinux:9.0
 RUN apk add --no-cache tzdata bash curl socat
```
wuhuizuo, line 2: O, `apk` is only for alpine. rocky linux is a RedHat like Linux. please use`yum` or `dnf` package manage tools.

jenkins/Dockerfile/release/debug-image/br
```diff
@@ -1,3 +1,3 @@
-FROM pingcap/alpine-glibc:alpine-3.14.6
+FROM rockylinux:9.0
 COPY zoneinfo.zip /usr/local/go/lib/time/zoneinfo.zip
```
wuhuizuo, line 2: Still need it now?
ref: https://github.com/golang/go/issues/34188

PR #1554, opened 2022-11-18
feat: run test in group

purelind:
run tests in group, case grouping via shell scripts to avoid frequent modification of ci scripts

test link  https://do.pingcap.net/jenkins/blue/organizations/jenkins/tikv%2Fmigration%2Fpull_integration_test/detail/pull_integration_test/18/pipeline/46


pingyu (approved):
LGTM~

PR #1561, opened 2022-11-21
feat(dm): only run test when match some files changed

purelind:
run compatibility tests only when some dm path files change.

replay test link
https://ci2.pingcap.net/job/dm_ghpr_compatibility_test/8802/console
https://ci2.pingcap.net/job/dm_ghpr_compatibility_test/8803/console


Ehco1996 (approved):
LGTM

PR #1588, opened 2022-11-28
fix: multibrach pd use branch name as version

lijie0123:
# Why
this is to fix  the issue that pd has version master rather than v6.5.0-alpha

# How
this PR just invoke build-common, remove inline logic for x86

# Hold, please fill why the path is like that


purelind (approved):

wuhuizuo:
The work is paused long time.

wuhuizuo:
The work is paused long time.

PR #1596, opened 2022-11-30
feat: make download file from fileserver more stable

purelind:
* add retry to fix unexpected error when download file from fileserver


purelind:
replay
- [x]  tidb_ghpr_common_task  https://ci.pingcap.net/blue/organizations/jenkins/tidb_ghpr_common_test/detail/tidb_ghpr_common_test/12294/pipeline/22/

PR #1605, opened 2022-12-02
feat: adjust internal goproxy

purelind:
adjust goproxy to avoid go mod download timeouts caused by the two network


purelind:
replay
- [x] https://ci.pingcap.net/blue/organizations/jenkins/tidb_ghpr_common_test/detail/tidb_ghpr_common_test/12373/pipeline

PR #1629, opened 2022-12-14
fix(pipelines/pingcap/tidb): fix concurrent go build cache conflicts in single container

wuhuizuo:
the speed looks faster then before


purelind (approved):
/lgtm

PR #1638, opened 2022-12-19
fix: add GOPROXY for tiflow docker build

lijie0123:
# WHY
build tiflow usually report connection reset when connectting to goproxy
# How
Add GOPROXY in Dockerfile
Signed-off-by: lijie <lijie@pingcap.com>


purelind (approved):
/lgtm

PR #1715, opened 2023-01-18
feat(libraries/tipipeline/vars): implement multi pull requests merging before build

wuhuizuo:
Refactor the share functions and the Jenkins jobs and pipeline scripts to support pre merge for multi pull requests together.

- when the `.pulls` in job refs is empty, only checkout the base commit of pull request.
- when the `.pulls` in job refs is not empty, merge the PR head branches to base. It's implemented in Jenkins Git plugin, but currently it will fail when size of `.pulls` greater then `1`, it will not make the CI wrong(prow will reduce the batch size to trigger CI job).

I will create another pull request to refactor lines like: https://github.com/PingCAP-QE/ci/pull/1715/files#diff-2ec31602787b9f78109d406f68793c25b46975c1ccf60392c646f117bbd09085R9

Only need `JOB_SPEC` param, so no need to fetch job spec from prow again.


wuhuizuo:
/hold
Need to test it.

wuhuizuo:
Currently multi PR pre merge to target branch will fail in Jenkins checkout, it should be a bug.

But the refactor is OK and tested for single pull requests. I think we should switch to `TektonCD` quickly.

PR #1716, opened 2023-01-18
fix(pipelines/pingcap): increase timeout for checkout with multi repos

wuhuizuo:
from 5 minutes to 10 minutes


purelind (approved):
/lgtm

PR #1744, opened 2023-02-03
chore: offline package not upload aws s3

purelind:
The `download.pingcap.org` domestic and foreign DNS are currently pointing to the qiniu cloud, so in the future do not need to upload to aws s3 offline package.


jenkins/pipelines/cd/tiup/tiup-package-offline-mirror-dm.groovy
```diff
@@ -0,0 +1,70 @@
+if(DEBUG_MODE == "true"){
```
wuhuizuo, line 1: The file is about the the motivation？

lijie0123: the file is added to CI from jenkins job script

lijie0123 (approved):

PR #1746, opened 2023-02-06
feat: speed up go mod download

lijie0123:
## Why
From 12min to 1min
image

image

## How
add go proxy in china


purelind (approved):
/lgtm

PR #1759, opened 2023-02-08
feat(pipelines/pingcap/tidb/latest): make pod in check2 job be a guaranteed pod

wuhuizuo:
request == limit = 8 core + 32 GB(from 16GB)


purelind (approved):
/lgtm

PR #1767, opened 2023-02-10
fix: v6.6.0 use rocky

lijie0123:
## Why
6.6.0 RC problem

## How:
- release check only main image, no need to check rocky
- sync more image for rokcy

Signed-off-by: lijie <lijie@pingcap.com>


purelind (approved):
/lgtm

PR #1815, opened 2023-02-20
mysql-test use bazel image

hawkingrei:


wuhuizuo:
already golang 1.20, and it has conflicts.

PR #1816, opened 2023-02-20
fix: 6.6.0 rocky check

lijie0123:
## Why
- rocky will use docker >= 20
- rocky lightning remove br


purelind:
> ## Why
> * rocky will use docker >= 20
> * rocky lightning remove br

alpine lightning still containes br, alpine image release check whether need check br

purelind (approved):
/lgtm

PR #1820, opened 2023-02-21
chore(pingcap/tiflow): tiflow upgrade to go1.20

purelind:
tiflow ci on master upgrade to go1.20.
* add go1.20 on old pipelines
* split pipelines for latest branch because goversion diverse from v7.0.0


purelind:
releated to https://github.com/pingcap/tiflow/pull/8328

PR #1821, opened 2023-02-21
chore(tikv/pd): split pd jobs for go1.20

purelind:
pd master go version will upgrade to go1.20, so we split pipelines from master.


purelind:
releated to https://github.com/tikv/pd/pull/6032

PR #1847, opened 2023-02-27
feat: support forked repo

lijie0123:
Why:
- to support forked repo
- for enterprise-plugin TIDB_HASH can accept repo name before sha1 to accomplish plugin with forked repo

### DIFF
--- a/jenkins/pipelines/cd/atom-jobs/build-common.groovy
+++ b/jenkins/pipelines/cd/atom-jobs/build-common.groovy
@@ -368,7 +368,7 @@ def checkoutCode() {
                                                 refspec      : specRef,
                                                 url          : tidb_repo]]]
             }
-            sh "git reset --hard ${TIDB_HASH}"
+            sh 'test -z "$(git status --porcelain)"'
         }
     }
 }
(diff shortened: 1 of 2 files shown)


jenkins/pipelines/cd/atom-jobs/build-common.groovy
```diff
@@ -352,8 +366,9 @@ def checkoutCode() {
                                         [$class: 'CleanBeforeCheckout']], submoduleCfg: [],
                             userRemoteConfigs: [[credentialsId: 'github-sre-bot-ssh',
                                                 refspec      : specRef,
-                                                url          : 'git@github.com:pingcap/tidb.git']]]
+                                                url          : tidb_repo]]]
             }
+            sh "git reset --hard ${TIDB_HASH}"
```
wuhuizuo, line 371: why need another git cli step to reset?

purelind (commented):
1. I think the current pipeline is already too complex and continuing to add new parameters 
which would make people confused may not be a good choice.
2. Support fork repo is a necessary feature or not, the current go version depends on tag or branch to determine, support fork reop build does not guarantee the use of normal go version

lijie0123:
1. It is pain to add more logic to this big script, but  the customer need forked repo because the cloud use forked repo , and it is better to be online before March
2. go version is selected by version prior than branch

purelind:
> 1. It is pain to add more logic to this big script, but  the customer need forked repo because the cloud use forked repo , and it is better to be online before March
> 2. go version is selected by version prior than branch, and version must be given

OK, LGTM, it will take some time to refactor the build-common.

purelind (approved):
/lgtm

PR #1850, opened 2023-02-28
chore: switch release-6.x pipelines of tidb-test to ksyun

purelind:
* Enable report on pipelines of ksyun.
* Disable trigger for pipelines on idc jenkins.


purelind:
Tested on release-6.1 and release-6.5
image

PR #1854, opened 2023-03-01
fix: use wget instead of curl

lijie0123:
## Why
- curl fileserver often fail with: curl: (18) transfer closed with 557466155 bytes remaining to read


purelind (approved):
/lgtm

PR #1857, opened 2023-03-01
feat: support failpoint in devbuild

lijie0123:
Why:
- sometimes we need failpoint
How:
- add a features fields, it can contains failpoint option


purelind (approved):
/lgtm

PR #1865, opened 2023-03-03
fix(security): use fix centos to repleace centos:8

lijie0123:
pingcap/centos-stream:8 is just centos:8 updated some packages


purelind (approved):
/lgtm

PR #1869, opened 2023-03-07
feat: upload enterprise plugin to fixed path

lijie0123:
Why:
- old-path is very long and complex
- jenkinsfile is not git managed


purelind (approved):
/lgtm

PR #1876, opened 2023-03-08
feat: add jenkins image syncer by pod

lijie0123:
Why:
- job is not managed by git
- old job use virtual-machine, which has been removed


purelind (approved):
/lgtm

PR #1878, opened 2023-03-08
fix(pingcap/tidb-test): tidb test add common ci fix

purelind:
* Correct s3 cache path 
* Enable report fot integration-common-test


purelind:
Tested https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb-test%2Fghpr_integration_common_test/detail/ghpr_integration_common_test/10/pipeline/

PR #1887, opened 2023-03-10
feat: enable WITH_RACE when build pd failpoint

purelind:
Enable WITH_RACE when build pd failpont binary.


lijie0123:
better not mixin race and failpoint

PR #1904, opened 2023-03-17
Build storage_consumer for ticdc tests

0xPoe:
We use it in some tests so we need to build it and upload it to CI. See: https://github.com/pingcap/tiflow/pull/8566


jenkins/pipelines/ci/ticdc/integration_test_common.groovy
```diff
@@ -89,6 +89,7 @@ def prepare_binaries() {
                             GOPATH=\$GOPATH:${ws}/go PATH=\$GOPATH/bin:${ws}/go/bin:\$PATH make cdc
                             GOPATH=\$GOPATH:${ws}/go PATH=\$GOPATH/bin:${ws}/go/bin:\$PATH make integration_test_build
                             GOPATH=\$GOPATH:${ws}/go PATH=\$GOPATH/bin:${ws}/go/bin:\$PATH make kafka_consumer
+                            GOPATH=\$GOPATH:${ws}/go PATH=\$GOPATH/bin:${ws}/go/bin:\$PATH make storage_consumer
```
purelind, line 92: Currently different branches share groovy, do you need to take into account that some branches do not exist `make storage_consumer`

purelind: How about let `make integration_test_build` contains `make storage_consumer `,  this will not break different branch pipelines.

PR #1914, opened 2023-03-20
fix(scrpts/plugins): fix report-flaky-cases about report time

wuhuizuo:
- enable report in release-6.5 branch
- fix report time col NULL problem.


purelind (approved):
/lgtm

PR #1924, opened 2023-03-22
ci: skip engine integration test if only pkg is changed

amyangfei:
as title said.


amyangfei:
PTAL @wuhuizuo @lijie0123

purelind (approved):
/lgtm

PR #1935, opened 2023-03-24
feat: Test the triggering of jobs based on labels.

purelind:
Test the triggering of jobs based on labels.


wuhuizuo:
prow not support label matching for trigger

PR #1942, opened 2023-03-28
fix(scripts/plugins): fix parallel sql inserting issue

wuhuizuo:


wuhuizuo:
Also it's deployed.

PR #1945, opened 2023-03-29
feat: will upload plugin to proper file-server path for non LTS

lijie0123:
## Why:
old non-LTS internal plugin path is ugly

## How:
add an option to specify whether it is LTS, non LTS will not push to public, but still push to internal


purelind (approved):
/lgtm

PR #1950, opened 2023-03-30
feat(tiflow): support skip ci if only change docs

purelind:
support skip ci if only change docs.

Test passed  https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflow%2Fpull_cdc_integration_test/detail/pull_cdc_integration_test/47/pipeline/

image


CharlesCheung96 (approved):

wuhuizuo:
why not skip the triggering when only change docs?

PR #1955, opened 2023-03-31
feat: add ad-hoc pipeline just for push br-cse to tiup stagine

lijie0123:


wuhuizuo:
Now we use the new CD flows, no need to merge it any more!

PR #1984, opened 2023-04-14
cdc: add G25 and G26 to cdc tests

CharlesCheung96:


purelind (commented):
Is it necessary to add two more test groups, which will consume more resources? It is recommended to add the new cases to the existing group with shorter execution time.

PR #2001, opened 2023-04-17
fix(gitee/pipelines/pingcap_enterprise):  add git ssh key in go build containers

wuhuizuo:
why: some repo's script will using git to clone or update, but we only set it in default container.


wuhuizuo:

------

### Commit 1
The pull request proposes a change in the git.groovy file to add SSH authentication for git clone or other operations. The change adds the capability to set the git SSH key in the Go build containers. The proposed change seems to target a known problem where some of the scripts use git to clone or update, but the SSH key was only set in the default container.

The change adds a line of code to the existing script and is well-documented. As there is no additional code added, there is no risk associated with it. However, a potential issue could arise if the SSH key is not set correctly, which could impact the build process. As long as the key is carefully set in secure containers, the impact should be minimal.

PR #2008, opened 2023-04-18
fix(jenkins/pipelines/ci/tiflash): reorder the steps in catch block

wuhuizuo:
Archive will failed, but we need the core log info, so reorder it.


wuhuizuo:

------

### Commit 1
The pull request fixes the order of steps when an error occurs during the Tiflash integration tests in a Jenkins pipeline. More specifically, the logging and archiving of the core log info are now done before throwing the error, which allows core information to be preserved even if the archiving step fails.

The patch only changes the order of two lines of code inside a catch block. No additional logic was added, and no existing code was removed.

A potential issue is that the order of the catch block steps was likely intentional, and this change could have unforeseen consequences. The new order may have unintended side effects, such as causing the build to hang or loop indefinitely. Another potential problem is that this change only addresses one specific scenario where the artifact archiving fails. There may be other scenarios where the logging and artifact archiving need to be handled differently. 

However, given that the changes are limited to a specific block of code and no other areas of the codebase are affected, the risk of unintended consequences should be low. It would still be worthwhile to test the change thoroughly to ensure that it does not introduce new issues.

PR #2010, opened 2023-04-19
fix(gitee/pipelines/pingcap_enterprise/tidb-enterprise-utilities): fix mysql tests

wuhuizuo:
try to fix and update notify


wuhuizuo:

------

### Commit 1
The pull request is a simple code change to four different files by a single developer. The change is to add Font Awesome icons to some text. The primary concern is that the patch does not include any information about what the changes really mean, so it's unclear why the Font Awesome icons were added. Additionally, it is not clear if the change was tested or why it was necessary. It would be helpful to have more detail in the commit message and a thorough testing report.

PR #2012, opened 2023-04-19
Update pr-verify.groovy

keeplearning20221:


wuhuizuo:

------

### Commit 1
The key change in this patch is the update to pr-verify.groovy file. Specifically, in the pipeline code, the branch name `REFS.base_ref` has been replaced with a hard-coded branch name `release-6.5.1`, thus changing the branch from which the code is being checked out. 

A potential problem with this patch is that the hard-coded branch name may cause issues in the future if the branch name changes or if the code needs to be checked out from a different branch. It may be better to use a variable or parameter to define the branch name. Additionally, it would be helpful to have more context about why this change is being made and to review the rest of the PR to ensure that this change does not break any other part of the codebase.

PR #2014, opened 2023-04-19
fix:  remove nfs on ci

purelind:
* Remove  useless nfs on ci pod
* Add todo for removing nfs dependency next time.

### DIFF
--- a/jenkins/pipelines/ci/pd/pd_ghpr_build.groovy
+++ b/jenkins/pipelines/ci/pd/pd_ghpr_build.groovy
@@ -70,8 +70,6 @@ def run_with_pod(Closure body) {
                     )
             ],
             volumes: [
-                    nfsVolume(mountPath: '/home/jenkins/agent/ci-cached-code-daily', serverAddress: '172.16.5.22',
-                            serverPath: '/mnt/ci.pingcap.net-nfs/git', readOnly: false),
                     emptyDirVolume(mountPath: '/tmp', memory: false),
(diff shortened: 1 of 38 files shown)


wuhuizuo:

------

This GitHub pull request includes three individual patches. The first two have no potential problems or errors. However, the third patch removes the NFS volume from the CI pipeline. This could lead to issues if data was previously being shared between different parts of the pipeline using NFS. It is unclear what alternatives are being used instead of NFS and whether they will be effective in replacing the removed volume.

## Details

### Commit 1
The patch changes the skip_report value from true to false for several jobs in the prow-postsubmits.yaml file in the pingcap/tidb/latest directory. This change enables mergeci reports on these jobs that were previously being skipped.

There are no potential problems of concern with this code change.

### Commit 2
The key change in this patch is modifying the `context` parameter from `wip-integration-jdbc-test` to `ci/integration-jdbc-test`. The change is made in the `jobs/pingcap/tidb/latest/prow-postsubmits.yaml` file. 

There are no potential problems identified in this patch.

### Commit 3
The key change in this pull request is the removal of the NFS volume on CI. There were several instances of `nfsVolume` being removed and replaced with `emptyDirVolume` or no volume at all. This will affect where data is stored during the CI process.

There may be potential problems with removing the NFS volume. The reason for using NFS may have been to share data between different parts of the CI pipeline. If this is the case, removing the NFS volume may lead to data being lost or not being shared properly. It is also unclear what the alternatives to NFS are and whether they will work effectively in place of the removed volume.

PR #2026, opened 2023-04-21
fix(tidb): master jobs not response to release-7.x

purelind:
master jobs not response to release-7.x

### DIFF
--- a/jobs/pingcap/tidb/latest/prow-presubmits.yaml
+++ b/jobs/pingcap/tidb/latest/prow-presubmits.yaml
@@ -10,7 +10,6 @@ presubmits:
       rerun_command: "/test build"
       branches:
         - ^master$
-        - ^release-7\.[1-9](\.\d+)?(-\d+)?(-v[\.\d]+)?$
         - ^feature[_/].+
     - name: pingcap/tidb/ghpr_check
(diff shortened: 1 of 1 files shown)


wuhuizuo:

------

### Commit 1
Key changes: 

- Removed regular expressions that specify the branches for which the jobs in the YAML file will be triggered. Specifically, the `^release-7\.[1-9](\.\d+)?(-\d+)?(-v[\.\d]+)?$` pattern was removed from all jobs' branch specifications. 

Potential problems:

- It's not clear from the commit message or the code changes what the purpose of this patch is, what issues it aims to address, and how it does so. The commit message simply says that it introduces a new feature of jobs responding to release-7.x, but the actual changes suggest that the feature is removed instead. This can be confusing and make it difficult for other developers to understand the reasoning behind the changes.
- The removal of the regular expressions that specify the branches for which the jobs will be triggered may have unintended consequences on the CI/CD pipeline for this project. It's not clear from the patch why these branches were removed and whether this change is intentional or a mistake. It's possible that removing these branches may cause some jobs to stop running or run on the wrong branches. Therefore, it's important to have a clear explanation of the reasoning behind this change and to thoroughly test the changes to make sure they work as intended.

PR #2027, opened 2023-04-21
feat(gitee): move out the jobs and pipeline files

wuhuizuo:
It was stored in other git repo at gitee.com.


purelind (approved):
/lgtm

PR #2030, opened 2023-04-26
cd: fix tidb-operator helm chart repo schema

csuzhangxc:
ref: https://github.com/pingcap/tidb-operator/issues/4984


wuhuizuo:

------

### Commit 1
The patch fixes the `tidb-operator` helm chart repository schema by changing the repository URL from `http` to `https`. This change is made in the `jenkins/pipelines/cd/tidb-operator.groovy` file.

The potential problem with this patch is that it may not be backward compatible with older versions of helm which do not support `https` schema. The patch author should confirm that the new schema is supported by all the helm versions that are in use in the project.

lijie0123 (approved):

PR #2032, opened 2023-05-04
fix: will print exception

lijie0123:
## Why
some build failed but we can't see any error message


jenkins/pipelines/cd/nightly/release-all-images-by-branch.groovy
```diff
@@ -740,6 +740,7 @@ try {
     }
 } catch (Exception e) {
     currentBuild.result = "FAILURE"
+    throw e
 } finally {
```
wuhuizuo, line 744: no need to add catch block, Only need finally block.
Ref: https://www.tutorialspoint.com/groovy/groovy_exception_handling.htm

PR #2065, opened 2023-05-12
chore: update default cluster for pod

purelind:
update default cluster for pod


wuhuizuo:
Is it ready? @purelind

PR #2069, opened 2023-05-14
fix(libraries/tipipeline/vars): clean untracked submodule's dirs

wuhuizuo:
Ref: #2068


libraries/tipipeline/vars/prow.groovy
```diff
@@ -50,6 +50,7 @@ def checkoutRefs(refs, timeout = 5, credentialsId = '', gitBaseUrl = 'https://gi
             echo "✅ Pre merged 🎉"
         fi
 
+        git clean -ffdx
```
purelind, line 53: Is this step necessary?

purelind (approved):
/lgtm

Issue #2112, opened 2023-06-02
Please do not using jenkins git plugin checkout for prow style jobs


wuhuizuo:
https://github.com/PingCAP-QE/ci/blob/5ea80488364787bb000f08544c2b0388172a7c54/pipelines/pingcap/tiflow/latest/pull_cdc_integration_kafka_test.groovy#L70C38-L85

It dose not block the issue in PR: pingcap/tiflow#8931

- single trigger mode: https://prow.tidb.net/?repo=pingcap%2Ftiflow&type=presubmit&pull=8931
- batch trigger mode: https://prow.tidb.net/?repo=pingcap%2Ftiflow&type=batch&pull=8931

It's fake passed in the batch triggered run since it only checked out the first PR's code.

PR #2114, opened 2023-06-02
fix(pipelines/pingcap/tiflow): fix checkout for prow style jobs

wuhuizuo:
Replace jenkins checkout with `prow.checkoutRef` to avoid the wrong checkouts for batch triggered runs.

Close issue #2112


wuhuizuo (commented):
Testing in: https://do.pingcap.net/jenkins/job/pingcap/job/tiflow/job/pull_cdc_integration_test/691/

It should failed since PR PingCAP/tiflow#8931 introduced the wrong script.

PR #2157, opened 2023-06-13
using bazel_importintotest to replace loaddatatest

D3Hunter:
see https://github.com/pingcap/tidb/pull/44602, we plan to rename bazel_loaddatatest to bazel_importintotest to avoid mixing with load data


purelind (approved):
/lgtm

PR #2168, opened 2023-06-15
fix: array is not support by bash in python image

wd0517:
error log: https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fmerged_integration_django_test/detail/merged_integration_django_test/19/pipeline/


purelind (approved):
/lgtm

PR #2266, opened 2023-07-13
feat: add lib checkout to only pull current pr

purelind:
Only checkout code of the current PR instead of all PRs, reducing unnecessary consumption of download bandwidth.

Add a checkoutV2 method to avoid unexpected errors affecting all related pipelines.When running for a period of time without any unexpected situations, replace the existing logic that uses checkout.

Notice: please clean lib cache in do jenkins after pr merged.


purelind:
reply test  https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb-test%2Fghpr_mysql_test/detail/ghpr_mysql_test/218/pipeline

PR #2267, opened 2023-07-13
fix(tidb-test): fix checkout tidb pr timeout

purelind:
The existing checkout method has a flaw where specifying the tidb pr causes a timeout due to checking out all PRs.


purelind:
wait https://github.com/PingCAP-QE/ci/pull/2266

PR #2279, opened 2023-07-18
feat(tiproxy): add two integration orm test

purelind:
Add two integration orm test.

Notice:
1. remove `wip/` from context after pipeline test passed


purelind:
After merged, test this in pr https://github.com/pingcap/TiProxy/pull/326

Issue #2277, opened 2023-07-17
Add prow OWNERS file to pingcap/tidb repo to control approving for important scopes


wuhuizuo:

wuhuizuo:
Add new commit to solve the parent dir option when no `.*` filter configurated.

Ref: 
- https://github.com/kubernetes/test-infra/pull/29970

PR #2280, opened 2023-07-18
feat(tidb): add notify presubmit job for tidb

wuhuizuo:
only notify when the monitor files changed.

Ref: #2277


purelind (approved):
/lgtm

PR #2287, opened 2023-07-19
remote hibernate test from TiProxy pipeline

djshow832:
- Remove `'make deploy-hibernatetest ARGS="-x"'` because it's unstable
- Add `'make deploy-tidbjdbctest ARGS="-x -m unique"'`


purelind (approved):
/lgtm

PR #2297, opened 2023-07-20
feat(tiflow): split tiflow storage test on v7.1.0

purelind:
split tiflow storage test on v7.1.0.


purelind:
Sadly to exclude strings containing 'v7.1.0', i will need to use a negative lookahead. Unfortunately, Go's standard regular expression package regexp does not support negative lookahead. I'll think about how to handle it again.

wuhuizuo (approved):
/lgtm
/approve

PR #2304, opened 2023-07-25
feat(tiproxy): tidb test add more tiproxy ci

purelind:
Add more tiproxy releated ci on tidb-test


pipelines/pingcap/tidb-test/latest/pull_tiproxy_ruby_orm_test.groovy
```diff
@@ -0,0 +1,127 @@
...
+                                }
+                            }
+                        }
+                        post{
+                            failure {
+                                script {
+                                    println "Test failed, archive the log"
+                                }
+                            }
+                        }
```
wuhuizuo, line 121: I think it's no need.

PR #2306, opened 2023-07-25
fix(tidb): adjust pipeline dependency cache

purelind:
Optimize pipeline cache logic.


purelind:
test passed  https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fmerged_integration_jdbc_test/detail/merged_integration_jdbc_test/1748/pipeline/260/

PR #2320, opened 2023-07-28
chore(tiflash): use specify node to reduce the probability of 5000 timeout issues

purelind:
use specify node to reduce the probability of 5000 timeout issues


purelind:
test link  https://ci.pingcap.net/job/tiflash-build-common/28299/console

PR #2317, opened 2023-07-27
feat: update GCR tag by plan

lijie0123:
# Why:
fix [#2314](https://github.com/PingCAP-QE/ci/issues/2314)
according to `GCR 镜像命名RFC `

PR #2321, opened 2023-07-28
fix: recover pipeline start at

lijie0123:
Why:
- fix PipelineStartAt unintended removed introduced by [#2317](https://github.com/PingCAP-QE/ci/pull/2317)


purelind (approved):
/lgtm

PR #2323, opened 2023-07-30
fix(tidb-test): fix s3 cache src code error

purelind:
fix s3 cache src code error.


purelind:
test passed https://do.pingcap.net/jenkins/job/pingcap/job/tidb-test/job/ghpr_common_test/48/console

PR #2350, opened 2023-08-10
feat(pd): update go version to 1.21

purelind:
pd master update go version to 1.21.

Ref  https://github.com/tikv/pd/issues/6923


purelind:
Ref  https://github.com/tikv/pd/pull/6924

PR #2359, opened 2023-08-16
fix(build): fix build dm multi branch fail

purelind:
* fix build dm multi-branch task fail caused by changes in file paths.
* remove arm64 binary build.


purelind:
test passed

PR #2367, opened 2023-08-21
fix(build): add nodeselector

purelind:
add nodeselector to avoid accidental scheduling to arm64 nodes.

### DIFF
--- a/jenkins/pipelines/cd/tiflow-operator-docker.groovy
+++ b/jenkins/pipelines/cd/tiflow-operator-docker.groovy
@@ -46,6 +46,7 @@ pipeline {
             agent {
                 kubernetes {
                     yaml podYaml
+                    nodeSelector "kubernetes.io/arch=amd64"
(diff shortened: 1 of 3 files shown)


xhebox (approved):

jenkins/pipelines/cd/tiflow-operator-docker.groovy
```diff
@@ -46,6 +46,7 @@ pipeline {
             agent {
                 kubernetes {
                     yaml podYaml
+                    nodeSelector "kubernetes.io/arch=amd64"
```
lijie0123, line 49: It seems not nessarary, because docker is multi-platform

purelind: We prefer to prioritize the use of amd64 nodes because there are fewer arm64 resources available.

lijie0123 (approved):

PR #2394, opened 2023-08-30
fix(tidb): archive the mysql test fail log

purelind:
Correct the log path for mysql test log archive.


purelind:
tested https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fghpr_mysql_test/detail/ghpr_mysql_test/28258/artifacts

PR #2399, opened 2023-08-31
feat: arm use pod rather than bare machine

lijie0123:
Why:
- pod is more reliable and reproducible
- we are migrating IDC, in order to keep service

### DIFF
--- a/jenkins/pipelines/cd/atom-jobs/build-common.groovy
+++ b/jenkins/pipelines/cd/atom-jobs/build-common.groovy
@@ -261,7 +261,7 @@ def containerLabel = "golang"
 def binPath = ""
 def useArmPod = false
 
-if (params.ARCH == "arm64" && params.PRODUCT in ["tidb", "enterprise-plugin", "tics"]) {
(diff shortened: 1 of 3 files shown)


jenkins/pipelines/cd/atom-jobs/build-common.groovy
```diff
@@ -1000,7 +999,7 @@ def run_with_arm_go_pod(Closure body) {
 try {
     stage("Build ${PRODUCT}") {
         if (!ifFileCacheExists()) { 
-            if (params.PRODUCT in ["tidb", "enterprise-plugin"] && params.ARCH == "arm64" &&  params.OS == "linux") {
+            if ( params.ARCH == "arm64" &&  params.OS == "linux") {
```
purelind, line 1002: This needs to be further modified here, excluding the tiflash arm64 build because tiflash uses a separate image.

purelind (approved):
/lgtm

purelind (commented):
/lgtm

purelind (approved):
/lgtm

PR #2400, opened 2023-08-31
fix: print error at exception

lijie0123:
# Why:
- fix the problem when pipeline failed but we do not know why


purelind (approved):
/lgtm

PR #2409, opened 2023-09-05
fix: force tikv linux/arm64 use 64k pagesize

lijie0123:
## Why:
- jemalloc in tikv use host pagesize, but 4k page binary can not run 64k kernel
- larger page size is compatible with lower page size: https://github.com/jemalloc/jemalloc/issues/467
- there is no any pitfall if builder is 64k. https://github.com/jemalloc/jemalloc/blob/b71da25b8a12c2c3f0c10b0811d15a61980186e8/configure.ac#L1805

## How:
- force tikv on linux/arm64 use 64k page size

## Tested:
- [x] builder 64k direct built binary can run on 4k machines
- [x] builder 4k with this PR built binary can run on both 4k and 64k machines


wuhuizuo (approved):
/lgtm

PR #2420, opened 2023-09-07
fix: enterprise plugin build

lijie0123:
## Why:
fix build enterprise plugin error after using arm pod

## How:
remove inner container directive


purelind (approved):
/lgtm

PR #2427, opened 2023-09-08
fix: hotfix will fail if it contains special features

lijie0123:
## Why:
- someone just give failpoint image as hotfix to customer

## How:
- just fail to avoid this situation before I have better solution


purelind (approved):
/lgtm

PR #2456, opened 2023-09-21
fix: rename pipeline name of mysql-conector-test

purelind:
rename pipeline mysql-conector-test to mysql-client-test.


bb7133 (approved):
LGTM

Issue #2465, opened 2023-09-22
stop using tiup:nightly after tiup release


lijie0123:
- to fix https://github.com/pingcap/tiup/issues/2268, tiup fix the bug only on nightly version
- after the statble version is released, stop using tiup nightly

PR #2476, opened 2023-09-26
chore: use tiup stable version

lijie0123:
# Why:
- tiup has release stable version to solve timeout problem

this PR will close: #2465


purelind (approved):
/lgtm

PR #2499, opened 2023-10-08
chore(pingcap/tidb): compatible with the coming package refactor contribution

wuhuizuo:
## Why
In the coming pull requests of `pingcap/tidb` repo, the folder layout will be changed, but will keep make tasks persistent.
we need use make to build rather than native go build commands to make CI jobs be compatible.


wuhuizuo:
Currently we only change CI jobs on trunk branches. other branches will be later.

purelind (approved):
/lgtm

PR #2503, opened 2023-10-10
fix: fix the typeorm scripts

Mini256:
The bash interpreter in the `hub.pingcap.net/ee/ci/base` image doesn't support the array feature, and @purelind provides a way to work around it.

cc: @wuhuizuo


purelind (approved):
/lgtm

PR #2505, opened 2023-10-10
fix(enterprise-plugin): correct enterprise-plugin path

purelind:
go.mod in enterprise-plugin used relative path.

test passed  https://ci.pingcap.net/blue/organizations/jenkins/tidb_ghpr_build/detail/tidb_ghpr_build/81578/pipeline/84/


wuhuizuo (approved):
/lgtm
/approve

Thanks @purelind , I introduced the issue since it's not use go module in `v5.4.x`

PR #2517, opened 2023-10-16
fix(tidb): fix cache path and cache key

purelind:
* fix cache path and cache key
 * only run with tikv store so we remove axis name TEST_STORE


purelind:
test link https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fmerged_integration_nodejs_test/detail/merged_integration_nodejs_test/46/pipeline

PR #2532, opened 2023-10-17
fix: tidb tools on arm

lijie0123:
## Why:
- tidb-tools on arm will not auto use latest go, which is a bug because tidb-tools always use master branch
- go version selection is hard to understand

## How:
- for tidb-tools>=5.3, directly select latest go version
- refract go version selection code
  - use switch instead of multiple if, with re-ordering
  - for non tidb-tools, no difference
  - for tidb-tools<5.3, no difference
  - for tidb-tools >= 5.3, no difference


purelind (approved):
/lgtm

PR #2566, opened 2023-10-27
feat(tidb): set pull-mysql-client-test as required lable on master branch

purelind:
Set `pull-mysql-client-tes`t as required label on master branch. After this PR is merged, each PR in the tidb master branch will be required to pass this test before merging.


bb7133 (approved):
LGTM

PR #2567, opened 2023-10-27
feat(pd): add integration-test on pd verifyCI

purelind:
## Why
In order to discover more issues before merging the PR, and avoid merging problematic PRs into the master branch, which may affect end-to-end testing of other components.

## How
To add an integraion-test-with-tidb on `tikv/pd` verifyCI  of master branch  (before PR merge). 

## Note
After the new pipeline tests passwd, we need to update the following two presubmit configurations.
*  `skip_report: true` --> `skip_report: false`
*  `optional: true ` remove


purelind:
After communicating with the relevant teams, we have decided to adopt a non-intrusive approach. Therefore, we will close this PR.

PR #2570, opened 2023-10-30
split addindex real-tikv test into multiple directory

ywqzzy:
currently, test in real-tikv-test is run in serial, if there're too many test in one directory, test will run slow and delay pr merge, or even timeout

so we add 4 addindextestX into real tikv test, we will split the test later, to avoid test timeout as we will add more test later.


purelind (approved):
/lgtm

Issue #2021, opened 2023-04-20
Failed to extract logs when tiflash integration test failed


hongyunyan:
When tiflash's integration test failed, ci can't successfully extract log files, so we can not get the enough information about the process, which make debug toughly. 
such as https://ci.pingcap.net/blue/organizations/jenkins/tiflash-ghpr-integration-tests/detail/tiflash-ghpr-integration-tests/12480/pipeline

<img width="1324" alt="image" src="https://user-images.githubusercontent.com/2574558/233249248-d8c7b74c-7d2e-4d1a-96be-cc30dbb037ec.png">
<img width="648" alt="image" src="https://user-images.githubusercontent.com/2574558/233249303-e2e72dff-358c-4f30-860b-3a59c4d1d372.png">

windtalker:
I also encounter this problem :(

JaySon-Huang:
+1

PR #2580, opened 2023-11-02
fix(tiflash): fix archive log failed 

purelind:
Close https://github.com/PingCAP-QE/ci/issues/2021
## Why
Fix archive log failed after some integration test failed due to missing file read and write permission in containers. 
The default user in image hub.pingcap.net/jenkins/docker:build-essential-java is `root`, but the jenkins try to archive log using `UID 1000`  within `archiveArtifacts(artifacts: "${name}-logs.tar.gz", allowEmptyArchive: true)`
image

## How
We change the ownship of log file to user jenkins (its UID is 1000)  before  archive logs.


purelind:
Test arhcive logs link  https://ci.pingcap.net/blue/organizations/jenkins/tiflash-ghpr-integration-tests/detail/tiflash-ghpr-integration-tests/14916/pipeline/

PR #2588, opened 2023-11-03
chore: update etcd from 3.3.10 to 3.4.21(since 6.5)

lijie0123:
Why:
- 3.3.10 is too old, even tidb 5.x use 3.4.3
- from tidb 6.5, use 3.4.21
- 3.4.21 copatible 3.4.3 by semantic versioning

How:
- use official etcd release package for linux, darwin/amd64
- build for darwin/arm64 with dockerfile:
```Dockerfile
FROM golang:1.19
ENV GOOS=darwin\
        GOARCH=arm64
ADD . /go/src/github.com/coreos/etcd
WORKDIR /go/src/github.com/coreos/etcd
RUN  ./build
```


wuhuizuo:
- [ ] need to update file in https://github.com/pingcap-qe/artifacts

PR #2600, opened 2023-11-08
chore: remove tispark and spark since v7.5

lijie0123:
Why:
- remove spark and tispark since v7.5


wuhuizuo (approved):
OK, it is approved by email.

PR #2602, opened 2023-11-09
feat: build tidb server without race

purelind:
Build tidb-server which not enable race in mysql-test pipelines.

Comparison of test time before and after starting the race.
* race  about 30 minutes
* without abount 12 minutes

Now, due to the presence of cache binary logic in pipelines, even though the build command is enabled race `WITH_RACE=1 make server`, it is actually using a tidb-server without race enabled for testing.


bb7133:
> Comparison of test time before and after starting the race.

> race about 12 minutes
> without abount 30minutes

Why it takes ~30mins without race?

purelind:
> > Comparison of test time before and after starting the race.
> 
> > race about 12 minutes
> > without abount 30minutes
> 
> Why it takes ~30mins without race?

Sorry, there was a mistake here. Only take 12minutes when not enable race.

Issue #2193, opened 2023-06-29
tikv ut pipelines upload cache file failed but not exit


purelind:
The cache will be uploaded to the fileserver in front of pipeline. If there is a failure during the upload process without any exception termination, it will cause errors when downloading the cache in subsequent steps.

PR #2608, opened 2023-11-13
feat: optimize upload cache of tikv ut

purelind:
Close https://github.com/PingCAP-QE/ci/issues/2193
## Why
The existing upload requires asynchronous uploading of each UT test binary, which may result in some caches not being successfully uploaded. This will affect the subsequent testing.

## How
After optimization, we pack all the ut test binary files that need to be uploaded together and upload them to ks3. At the same time, we only use ks3util for uploading and downloading operations. This not only improves the stability of uploads but also reduces the time required for uploading and downloading.
Upload time before optimization: 50s
Upload time after optimization: 22s


purelind:
Test passed
1. ut with nexttest https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/39920/pipeline/10/
2. ut legacy  https://ci.pingcap.net/blue/organizations/jenkins/tikv_ghpr_test/detail/tikv_ghpr_test/39925/pipeline/

glorv (commented):
LGTM

PR #2614, opened 2023-11-14
do not release tikv-importer chart as it is deprecated

csuzhangxc:


lijie0123 (commented):
will it influence old version?

csuzhangxc:
> will it influence old version?

we do not want to release any version of tikv-importer anymore. (including v1.4.x and v1.5.x)

lijie0123 (approved):

PR #2629, opened 2023-11-23
feat(tiflash): add tiflash code cache

purelind:
Add tiflash code cache at the first checkout step. After adding the code cache, the time it takes to checkout the code decreased from 30s to 8s.

In case of special circumstances, such as GitHub's speed limit, due to the presence of code cache, it will be less likely to encounter checkout timeout issues.


purelind:
test link  https://ci.pingcap.net/blue/organizations/jenkins/tiflash-build-common/detail/tiflash-build-common/32202/pipeline/38


image

JaySon-Huang (approved):

PR #2632, opened 2023-11-27
feat(tidb/lightning): add pull lightning tests pipeline

purelind:
Add new pipeline for lightning test, support split lightning test from current pull_br_integration_test.


purelind:
relate to https://github.com/pingcap/tidb/pull/48921

PR #2642, opened 2023-11-29
fix(tidb): archive log for all unsuccessful task

purelind:
* Currently, only when the task fails will it be archived. If the task times out or fails unstably, no logs can be collected.
* The logs generated by parallel tasks will overlap with each other, so here we first use tar to package the logs and then archive them.


purelind:
replay test for timeout pipeline  https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fghpr_check2/detail/ghpr_check2/34412/artifacts

wuhuizuo (approved):

PR #2571, opened 2023-10-31
feat(script/plugins): support analyze fail reason of test case

wuhuizuo:
> currently it's implemented with go race analyzing.

## Tests

- [x] tested manually

```bash
# inputs from: https://do.pingcap.net/jenkins/job/pingcap/job/tidb/job/ghpr_unit_test/36206/artifact/
> ./analyze-go-test-from-bazel-output.sh bazel-test.log
Parse from local file: bazel-test.log
race detected case: TestAddIndexIngestGeneratedColumns
new flaky case: TestCursorFetchErrorInFetch
new flaky case: TestCursorFetchShouldSpill
Output files:
bazel-flaky-summaries.log                       bazel-go-test-problem-cases.json                bazel-target-output-L4631-5414.fail.log
bazel-go-test-index.log                         bazel-target-output-L3857-4621.fail.log         bazel-target-output-L520-3195.fail.race.log
```

## Results 

file `bazel-go-test-index.log`:
```log
520:==================== Test output for //pkg/ddl/ingest:ingest_test (shard 5 of 15):
521:=== RUN   TestAddIndexIngestGeneratedColumns
1790:WARNING: DATA RACE
2012:WARNING: DATA RACE
3185:    testing.go:1465: race detected during execution of test
3186:--- FAIL: TestAddIndexIngestGeneratedColumns (63.09s)
3188:    testing.go:1465: race detected during execution of test
3195:================================================================================
3857:==================== Test output for //pkg/server:server_test (shard 6 of 48):
3858:=== RUN   TestCursorFetchErrorInFetch
4614:--- FAIL: TestCursorFetchErrorInFetch (4.14s)
4621:================================================================================
4631:==================== Test output for //pkg/server:server_test (shard 5 of 48):
4632:=== RUN   TestCursorFetchShouldSpill
5407:--- FAIL: TestCursorFetchShouldSpill (2.22s)
5414:================================================================================
```

file `bazel-go-test-problem-cases.json`:
```json
{
  "//pkg/ddl/ingest:ingest_test": {
    "new_flaky": [
      {
        "name": "TestAddIndexIngestGeneratedColumns",
        "reason": "race"
      }
    ]
  },
  "//pkg/server:server_test": {
    "new_flaky": [
      {
        "name": "TestCursorFetchErrorInFetch",
        "reason": "unknow"
      },
      {
        "name": "TestCursorFetchShouldSpill",
        "reason": "unknow"
      }
    ]
  }
}

```

PR #2661, opened 2023-12-06
Revert "feat(script/plugins): support analyze fail reason of test case"

wuhuizuo:
Reverts PingCAP-QE/ci#2571


wuhuizuo:
fixed, no need it any more

PR #2663, opened 2023-12-06
feat(devbuild): support drainer and pump

lijie0123:
# Why:
- support drainer and pump as requested

# How:
- support docker image name differ from tiup

# Testing:
- re-run  for existing devbuild


purelind (approved):
/lgtm

PR #2664, opened 2023-12-07
feat: add 7.5 to lts

jayl1e:
Why:
- 7.5 is lts
- while in publishing 7.5.0, we already rerun with this change


purelind (approved):
/lgtm

PR #2669, opened 2023-12-11
feat: tiflow support fips branch pipelines

purelind:
tiflow support fips branch pipelines


prow-jobs/pingcap-tiflow-release-6.5-fips-presubmits.yaml
```diff
@@ -0,0 +1,60 @@
+# struct ref: https://pkg.go.dev/k8s.io/test-infra/prow/config#Presubmit
+presubmits:
```
wuhuizuo, line 2: it should be added to kustomization.yaml

PR #2685, opened 2023-12-15
feat: optimize job clone params

purelind:
* use shallow clone to reduce clone time.
* add affinity to schedule pod to x86 node.


purelind:
test on  https://prow.tidb.net/view/gs/prow-tidb-logs/logs/periodic-daily-tidb-unit-test/1735610189085675520

PR #2688, opened 2023-12-18
fix: solve yum install slow for centos 8

lijie0123:
# Why:
- yum install is very slow because centos has been deprecated.(usually 1 hour)

# How:
- use dockerfile from https://github.com/PingCAP-QE/artifacts/blob/main/dockerfiles/bases/tidb-base/release-6.5.Dockerfile
- sync it to dockerhub to avoid any network problem

### DIFF
--- a/jenkins/Dockerfile/release/linux-arm64/enterprise/tidb
+++ b/jenkins/Dockerfile/release/linux-arm64/enterprise/tidb
@@ -1,7 +1,4 @@
-FROM pingcap/centos-stream:8
-RUN set -e && \
-    dnf install bind-utils curl nmap-ncat -y && \
-    dnf clean all
+FROM pingcap/tidb-base:centos8
 COPY tidb-server /tidb-server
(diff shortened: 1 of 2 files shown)


jenkins/Dockerfile/release/linux-arm64/enterprise/tidb
```diff
@@ -1,7 +1,4 @@
-FROM pingcap/centos-stream:8
-RUN set -e && \
-    dnf install bind-utils curl nmap-ncat -y && \
-    dnf clean all
+FROM pingcap/tidb-base:centos8
```
purelind, line 1: Is all those packages `bind-utils curl nmap-ncat` included in image `pingcap/tidb-base:centos8`

lijie0123: yes, the dockerfile is in the PR content

purelind (approved):
/lgtm

PR #2693, opened 2023-12-21
debug: test prow job optional param

purelind:
test prow job optional param, I will manually trigger a task, and after this task fails, my PR merge should be blocked.


purelind:
this is debug pr , close it .

PR #2715, opened 2024-01-03
[DNM]Update pull_dm_integration_test.groovy

Benjamin2037:


purelind:
It is suggested to consider dispersing the tests in group11 among the existing groups instead of adding more groups. Because the execution time of other group tests is generally only 4 minutes.

purelind:
The issue has been resolved through this PR https://github.com/pingcap/tiflow/pull/10409.  I will close this PR.

PR #2729, opened 2024-01-08
fix(tikv/migration): fix missing volume

purelind:
* fix missing volume
* Use image with tini to avoid the problem of lsof getting stuck.


pingyu (approved):
LGTM~

PR #2733, opened 2024-01-08
chore: increase cpu request on build

purelind:
increase cpu request on build


pipelines/pingcap/tidb/latest/pod-ghpr_build.yaml
```diff
@@ -12,7 +12,7 @@ spec:
       resources:
         limits:
           memory: 32Gi
-          cpu: "8"
+          cpu: "16"
```
wuhuizuo, line 15: We should increase memory to 64Gi also

wuhuizuo (approved):

PR #2741, opened 2024-01-12
fix: operator git commit

lijie0123:
# Why:
- arm binary now use independent arm builder for fips
- we need export GIT_COMMIT every time or it would be overwrited by jenkins


purelind (approved):
/lgtm

PR #2749, opened 2024-01-16
tiproxy: add nodejs test

djshow832:
- Move prisma test and sequelize test to nodejs test, and add typeorm test to nodejs test, just like TiDB
- Update the pod of nodejs test to be the same as TiDB's. TiProxy's prisma test always fails: https://prow.tidb.net/log?job=pingcap/tiproxy/merged_integration_prisma_test&id=1746725158531371012


purelind (approved):
/lgtm

PR #2769, opened 2024-01-22
tikv-migration: collect logs with -maxdepth 2

pingyu:
### Changes

- Collect logs with `-maxdepth 2` to exclude WAL of TiKV & PD and reduce artifact size
- Add collect logs for `pull_integration_test` pipeline.


purelind (approved):
/lgtm

PR #2778, opened 2024-01-23
tikv-migration: increase pipeline timeout

pingyu:
### Changes

For tikv-migration CDC pipelines:

- Increase timeout from `40` to `65` minutes 
  - Original `40 minues` is not sufficient. See https://do.pingcap.net/jenkins/blue/organizations/jenkins/tikv%2Fmigration%2Fpull_integration_kafka_test/detail/pull_integration_kafka_test/28/pipeline/261/.
- Increase timeout of `Test` stage from `25` to `45` minutes


purelind (approved):
/lgtm

PR #2791, opened 2024-01-30
feat: disable nightly pipeline

lijie0123:
# Why:
- use new pipeline


wuhuizuo (approved):

wuhuizuo:
Thanks.

We switched to new CD flows.

PR #2804, opened 2024-02-18
fix: update checkout param to ensure submodule update correctly

purelind:
update checkout param to ensure submodule update correctly


purelind:
retest link
1. https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fmerged_build/detail/merged_build/431/pipeline/95/

PR #2809, opened 2024-02-21
fix(tiflow): fix get pr diff files not incomplete

purelind:
Fix get pr diff files not incomplete.  The paginated response of github api returns 30 files per page by default.


purelind:
replay test  https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflow%2Fpull_dm_compatibility_test/detail/pull_dm_compatibility_test/903/pipeline

PR #2825, opened 2024-02-29
chore: update prometheus version to 2.49.1

lijie0123:
# Why:
- v8.0.0 later will use prometheus version to 2.49.1, keep the same as https://github.com/PingCAP-QE/artifacts/blob/32b627f679746b7d1b521c9d630b99ed27ec6fa8/packages/packages.yaml.tmpl#L229

# Testing:
- tested via rerun https://cd.pingcap.net/blue/organizations/jenkins/prometheus-tiup-mirror-update/detail/prometheus-tiup-mirror-update/275/pipeline


purelind (approved):
/lgtm

PR #2830, opened 2024-03-06
fix: Using domestic sources to solve the failure of unstable network.

purelind:
Using domestic sources to solve the failure of unstable network.

image


purelind:
replay test link https://cd.pingcap.net/blue/organizations/jenkins/build-common/detail/build-common/296706/pipeline

lijie0123 (approved):

PR #2849, opened 2024-03-15
feat: stop push docker for rc

lijie0123:
Why:
- new RC build will publish docker for QA


wuhuizuo:
/hold

keep it when we require it.

PR #2867, opened 2024-03-22
fix: remove -c option for tiproxy mysql_test

djshow832:
Refer to https://github.com/PingCAP-QE/tidb-test/pull/2315: the `-c` is removed.


purelind (approved):
/lgtm

PR #2873, opened 2024-03-26
feat(scripts/artifacts): add qiniu object list and delete scripts

wuhuizuo:


purelind (approved):
/lgtm

PR #2892, opened 2024-04-08
fix: still use default src dir to reuse tartget cache

purelind:
still use default src dir to reuse tartget cache.


purelind:
test passed https://do.pingcap.net/jenkins/blue/organizations/jenkins/tikv%2Ftikv%2Fpull_unit_test/detail/pull_unit_test/4/pipeline

PR #2900, opened 2024-04-10
change trigger of lightning ci to include dependency pkg

D3Hunter:


lance6716 (approved):

lance6716 (commented):
How did you get the dependency lists? I use `go list -deps ./lightning/cmd/tidb-lightning/ | grep "github.com/pingcap/tidb"` and found there are more used packages, but many of them are not important

D3Hunter:
> How did you get the dependency lists? I use `go list -deps ./lightning/cmd/tidb-lightning/ | grep "github.com/pingcap/tidb"` and found there are more used packages, but many of them are not important

by search `pingcap/tidb/br/`. yes, some related to version/util/log are removed, it's too many

purelind (approved):

PR #2928, opened 2024-04-19
fix(br): use failpoint tidb-server instead

Leavrth:
Since https://github.com/pingcap/tidb/pull/52734, we need to use failpoint tidb-server for br integration test


purelind (approved):
/lgtm

PR #2932, opened 2024-04-22
increase ticdc pulsar integration test resource

sdojjy:
ticdc pulsar integration test failed frequently, but when we run it on local env, it never fail, so this PR increase ticdc pulsar integration test resource.


asddongmen (approved):

purelind (approved):
/lgtm

PR #2934, opened 2024-04-22
increase ticdc kafka integration test resource

sdojjy:
ticdc kafka integration test failed frequently, but when we ran it on local env, it never fail, so this PR increase ticdc kafka integration test resource.


purelind:
This is a temporary resource adjustment change. After confirming the issue, heavy testing needs to be migrated to avoid excessive resource occupation by individual pipelines.

purelind (approved):
/lgtm

PR #2937, opened 2024-04-23
fix(scripts/pingcap): fix script of adding version placeholder

wuhuizuo:


purelind (approved):
/lgtm

PR #2939, opened 2024-04-24
fix: upadte new ut on tikv 6.1

purelind:
upadte new ut on tikv 6.1


purelind:
test passed  https://do.pingcap.net/jenkins/job/tikv/job/tikv/job/release-6.1/job/pull_unit_test/18/

PR #2947, opened 2024-04-29
fix(tiflash): correct the dir to collect log

purelind:
correct the dir to collect log.


purelind:
test link  https://do.pingcap.net/jenkins/job/pingcap/job/tiflash/job/pull_integration_test/103/

PR #2986, opened 2024-06-07
feat: try to collect failure log after test timeout

purelind:
try to collect failure log after test timeout


purelind:
/hold

wait for relay test link  https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fpull_lightning_integration_test/detail/pull_lightning_integration_test/842/pipeline

PR #2995, opened 2024-06-13
fix(jenkins/pipelines/cd): keep same tags with inner repositories  on GCR

wuhuizuo:


purelind (approved):
/lgtm

PR #3035, opened 2024-07-29
feat(tools/list-rust-dependencies): add script to list rust dependencies to csv

wuhuizuo:

### DIFF
--- a/tools/list-rust-dependencies/list.ts
+++ b/tools/list-rust-dependencies/list.ts
@@ -1,0 +1,100 @@
+// run by cargo tree -f "{p} --  {l} -- {r}" | tee tree.full.log
+/*
+tikv-ctl v0.0.1 (/workspace/tikv/cmd/tikv-ctl) --  Apache-2.0 --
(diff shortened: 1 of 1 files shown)


wuhuizuo:
> > **I have already done a preliminary review for you, and I hope to help you do a better job.**
> 
> The changes in this pull request include adding a script to list Rust dependencies to CSV. The script parses the output of the `cargo tree` command and writes the dependencies to a CSV file. The script also includes a function to parse the depth of the dependency tree.
> 
> One potential problem is that the script assumes `cargo` is installed and available in the environment. This may not be the case on all systems or in all CI/CD environments. A suggestion for a fix could be to include a check for the presence of `cargo` and print an error message if it is not found.
> 
> Another suggestion for improvement would be to add some error handling to the script. For example, if the `cargo tree` command fails, the script should print an error message and exit with a non-zero status code.
> 
> Overall, the changes seem well-written and straightforward.

`cargo tree` is a built-in sub command of cargo, so do not worry it.

PR #3100, opened 2024-09-03
fix: not to set jnlp conf in pod file

purelind:
Not to set jnlp container config in pod yaml file.
* If we use the Groovy function fileExists, we must explicitly define the jnlp container because this function can only be executed in the jnlp container and cannot specify other containers.

### DIFF
--- a/pipelines/pingcap/tiflash/latest/pull_integration_test.groovy
+++ b/pipelines/pingcap/tiflash/latest/pull_integration_test.groovy
@@ -125,7 +125,7 @@ pipeline {
                 stage("Proxy-Cache") {
                     steps {
                         script {
-                            proxy_cache_ready = fileExists("/home/jenkins/agent/proxy-cache/${proxy_commit_hash}-amd64-linux-llvm")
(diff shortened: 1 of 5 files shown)


pipelines/pingcap/tiflash/latest/pull_integration_test.groovy
```diff
@@ -125,7 +125,7 @@ pipeline {
                 stage("Proxy-Cache") {
                     steps {
                         script {
-                            proxy_cache_ready = fileExists("/home/jenkins/agent/proxy-cache/${proxy_commit_hash}-amd64-linux-llvm")
+                            proxy_cache_ready = sh(script: "test -f /home/jenkins/agent/proxy-cache/${proxy_commit_hash}-amd64-linux-llvm && echo 'true' || echo 'false'", returnStdout: true).trim() == 'true'
                             println "proxy_cache_ready: ${proxy_cache_ready}"
```
wuhuizuo, line 129: Can we make it just be a `sh` step?

purelind: No, this variable needs to be passed to the subsequent steps for use.

wuhuizuo (approved):

PR #3108, opened 2024-09-06
RocksDB CI: use static link

v01dstar:
RocksDB switched to dynamic link. We may want to stick to static link for better portability,


purelind (approved):
/lgtm

PR #3129, opened 2024-09-19
Use RocksDB native parallel testing

v01dstar:


v01dstar:
`make check` fails in our CI, it requires python3. So, this change will be paused, until we migrate to new build env.

wuhuizuo:
No updates for a long time

PR #3171, opened 2024-10-11
fix(tiflash): user clang-format 15 after v7.4

purelind:
Related to this PR https://github.com/pingcap/tiflash/pull/7927/files#diff-a847e281c62ad2738f1616a96cd3174ca1fa5eba8134bf6f5d802a425c92c8b2R86-R90

For versions greater than or equal to 7.4 and less than 8.2, use clang-format 15.


JaySon-Huang (approved):
LGTM

PR #3189, opened 2024-10-29
feat(tiflow): update base image to use rockylinux8

purelind:
CentOS 7 has reached the end of its life cycle. Starting from version 8.4, we began using Rocky Linux 8 as the base image.


purelind:
pull_verify retest link:  https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflow%2Fghpr_verify/detail/ghpr_verify/2797/pipeline

PR #3191, opened 2024-10-29
fix(script/flow/rc): fix flow check script

wuhuizuo:
- skip check for tidb-binlog since v8.4.0.
- fix the shell sharbang.


purelind (approved):
/lgtm

PR #3207, opened 2024-11-06
feat(lib): use git shallow clone for lib cache

purelind:
* Use git shallow clone for lib cache to optimize the issue of incomplete library downloads caused by network quality.
* Disable the autorefresh lib for folder pingcap

### DIFF
--- a/jobs/pingcap-inc/aa_folder.groovy
+++ b/jobs/pingcap-inc/aa_folder.groovy
@@ -11,13 +11,13 @@ folder('pingcap-inc') {
                             scm {
                                 git {
                                     remote('https://github.com/PingCAP-QE/ci')
-                                }
-                                extensions {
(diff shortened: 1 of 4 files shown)


jobs/pingcap-inc/aa_folder.groovy
```diff
@@ -12,6 +12,13 @@ folder('pingcap-inc') {
                                 git {
                                     remote('https://github.com/PingCAP-QE/ci')
                                 }
+                                extensions {
+                                    cloneOptions {
+                                        depth(1)
+                                        shallow(true)
+                                        timeout(5)
+                                    } 
+                                }
```
wuhuizuo, line 21: whether it should be in the `git` block?

wuhuizuo (approved):

PR #3210, opened 2024-11-07
fix(tidb-engine-ext): link git and registry cargo cache

purelind:
Directly setting the CARGO_HOME variable will result in losing the default config and env configurations，so just creat link for git and registry of cargo cache.


CalvinNeo (approved):
lgtm

PR #3215, opened 2024-11-08
feat(script): add s3 object monitor clean script

purelind:
Add s3 object monitor clean script, The default S3 upload limit is 5GB, and some code caches may exceed this limit, causing the upload to fail.
* just monitor
```
deno run --allow-net --allow-env monitor-s3-object-size.ts \
  --path="path" \
  --threshold-mb=500 \
  --feishu-webhook="xxxx"
```
* monitor and clean
```shell
deno run --allow-net --allow-env monitor-s3-object-size.ts \
  --path="path" \
  --threshold-mb=500 \
  --feishu-webhook="xxxx" \
  --cleanup
```

### DIFF
--- a/scripts/plugins/monitor-s3-object-size.ts
+++ b/scripts/plugins/monitor-s3-object-size.ts
@@ -37,7 +37,7 @@ async function main() {
   const thresholdMB = args["threshold-mb"] || 100;
   const thresholdBytes = thresholdMB * 1024 * 1024;
 
-  // 飞书配置
+  // Feishu configuration
   const feishuConfig: FeishuConfig | undefined = args["feishu-webhook"]
     ? {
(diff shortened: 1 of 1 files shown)


scripts/plugins/monitor-s3-object-size.ts
```diff
@@ -0,0 +1,343 @@
...
+
+  // 飞书配置
+  const feishuConfig: FeishuConfig | undefined = args["feishu-webhook"]
+    ? {
+      webhook: args["feishu-webhook"] as string,
+      secret: args["feishu-secret"] as string | undefined,
+    }
+    : undefined;
+
+  // 清理配置
```
wuhuizuo, line 48: we prefer to comment codes with english words.

PR #3216, opened 2024-11-08
fix(tiflash): fix script path

Lloyd-Pottiger:
In https://github.com/pingcap/tiflash/pull/9587, we modify the directory name.


purelind:
duplicate with https://github.com/PingCAP-QE/ci/pull/3219

PR #3280, opened 2024-12-12
fix(tidb): prepare sqllogic test data in pod init

purelind:
After adjusting the CentOS 7 image to Rocky 8, in order to reduce the image size, we did not include the test files in the image but prepare them during the pod init phase.


purelind:
test link: https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftidb%2Fmerged_sqllogic_test/detail/merged_sqllogic_test/2686/pipeline/

PR #3303, opened 2024-12-25
fix(tiflow/dm): use tini image to avoid zombie process

purelind:
Use tini as init process in image to avoid zombie process after test kill some process.


purelind:
replay test link
1. https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflow%2Frelease-8.1%2Fpull_dm_integration_test/detail/pull_dm_integration_test/364/pipeline
2. https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflow%2Frelease-8.1%2Fpull_dm_integration_test/detail/pull_dm_integration_test/361/pipeline/131/

PR #3324, opened 2025-01-22
Remove mac test pipeline from storage repos

v01dstar:
TiKV's RocksDB's fork's CI disabled `mac` build anyway. And `mac` build is too slow in Titan's and rust-rocksdb's CI. So, disable `mac` build across storage repos' (RocksDB, Titan, rust-rocksdb) CI.

### DIFF
--- a/jenkins/pipelines/ci/rocksdb/rocksdb_ghpr_test.groovy
+++ b/jenkins/pipelines/ci/rocksdb/rocksdb_ghpr_test.groovy
@@ -119,15 +119,6 @@ parallel(
             test("", "db_block_cache_test", "", do_cache)
         }
     },
-    /*
-    mac: {
-        node("mac-i7") {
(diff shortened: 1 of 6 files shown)


glorv:
@Connor1996 PTAL

purelind:
Will this change increase the likelihood of TiKV failing to build on macOS?

Connor1996 (commented):
Many of us are still developing on mac, we better not disable it. Why tests run slow on macos?

purelind:
> Many of us are still developing on mac, we better not disable it. Why tests run slow on macos?

I think this is because the currently available Mac machine is the Mac Mini with an Intel i7 chip, which has relatively weak performance.

v01dstar:
How about using `mac-arm`?  @purelind is that a good option?

PR #3329, opened 2025-01-23
chore(tiflash): increase memory limit to avoid build oom

purelind:
The current memory limit is 32G, and during the tiflash compilation process, there are instances of OOM occurring on the master branch code.  Maybe recent changes of tiflash have led to a need for more memory.


JaySon-Huang (approved):

CalvinNeo (approved):
lgtm

PR #3432, opened 2025-03-20
feat(pipelines): Add proxy cache upload logic on master merged-build pipeline

purelind:
- Implemented logic to upload proxy cache only if the commit hash is valid and the refresh is enabled.
- Added checks to skip upload if the cache already exists or if the proxy library is not found.
- Updated messages to provide clear feedback on the upload process.


purelind:
retest link https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflash%2Fmerged_build/detail/merged_build/173/pipeline

JaySon-Huang (approved):
LGTM

PR #3435, opened 2025-03-24
feat(prow-jobs): add periodic job to crawl Git commits data

wuhuizuo:


wuhuizuo:
/hold

some secrets should be updated.

PR #3441, opened 2025-03-28
feat(ticdc): add net-tool cpu from 100m to 300m for kafka light pipelines

wlwilliamx:
In the monitoring of Kafka Light Pipelines, the CPU throttling metric for net-tool is consistently around 60%. Therefore, it is necessary to increase the CPU allocation for net-tool. This PR increases net-tool’s CPU from 100m to 300m.


purelind (approved):
/lgtm

PR #3485, opened 2025-04-21
feat(pingcap-inc/tici): update Rust build image to TiKV builder

wuhuizuo:
Remove manual protobuf installation as it's included in the TiKV builder image.

The image also contains CMake v3 tool.


purelind (approved):
/lgtm

PR #3494, opened 2025-04-27
fix(prow-jobs/pingcap-inc/tici): enable devtoolset-10 

wuhuizuo:
And increase resource limits to 2 CPU + 8Gi RAW


wshwsh12 (approved):

purelind (approved):
/lgtm

PR #3588, opened 2025-06-16
chore(prow-jobs): disable and remove presubmits for release-6.1 branches

wuhuizuo:
v6.1.x was EOS at 2025-06-13, so we will not trigger CI jobs for `release-6.1` branches.


wuhuizuo:
We should support the hot-fix on v6.1.x until it end of life.

PR #3647, opened 2025-07-15
chore(devbuild): add retry on package-tiup

purelind:
Add retry on package-tiup to avoid failures caused by network issues when uploading the tiup package.


purelind:
replay test passed  https://cd.pingcap.net/blue/organizations/jenkins/devbuild/detail/devbuild/8240/pipeline/90

PR #3667, opened 2025-08-01
fix go -race in tidb operator e2e test

liubog2008:
fix go -race in tidb operator e2e test


purelind:
This type of variable is recommended to be placed in the makefile, rather than increasing the complexity of the pipeline.

purelind (approved):
/lgtm

PR #3691, opened 2025-08-08
feat(pipelines/tidb): update pull-error-log-review job configuration of tidb

purelind:
- Changed the `always_run` property to `true` for the `pull-error-log-review` job, ensuring it runs on every trigger.
- Added `skip_report: true` to prevent reporting on this job, streamlining the CI process.

Enable automatic triggering. To verify the effectiveness of the matching rules, please note that this task currently has status reporting turned off, so task cost failures will not be reflected on the PR GitHub page and will not affect PR merging.

### DIFF
--- a/prow-jobs/pingcap/tidb/latest-presubmits.yaml
+++ b/prow-jobs/pingcap/tidb/latest-presubmits.yaml
@@ -206,7 +206,8 @@ presubmits:
       decoration_config:
         timeout: 20m
         skip_cloning: true
-      always_run: false
+      always_run: true
+      skip_report: true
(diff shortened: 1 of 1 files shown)


prow-jobs/pingcap/tidb/latest-presubmits.yaml
```diff
@@ -206,7 +206,8 @@ presubmits:
       decoration_config:
         timeout: 20m
         skip_cloning: true
-      always_run: false
+      always_run: true
```
wuhuizuo, line 209: like the build job, use skip_if_only_changed.

wuhuizuo (approved):

PR #3692, opened 2025-08-08
fix(jenkins/pipelines/cd): use new base images since v6.5.12

wuhuizuo:
This pull request introduces a semantic version comparison function to the `dev-build.groovy` pipeline script, improving the accuracy of version checks when selecting Dockerfiles. The main change replaces a simple string comparison with a proper semantic version comparator, ensuring correct ordering for versions with pre-release and patch information.

Version comparison improvements:

* Added a `semverCompare` closure to accurately compare semantic version strings, including handling of pre-release and patch metadata.
* Updated the Dockerfile selection logic to use `semverCompare` instead of a string comparison, ensuring correct behavior for versions such as `v6.5.12` and above.

### DIFF
--- a/jenkins/pipelines/cd/dev-build.groovy
+++ b/jenkins/pipelines/cd/dev-build.groovy
@@ -60,12 +60,39 @@ def ShortImageRepoForGCR = ''
 def NeedEnterprisePlugin = false
 def PrintedVersion = ''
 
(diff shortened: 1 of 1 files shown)


purelind (approved):
/lgtm

wuhuizuo:
/unhold

I appended a commit 03d32297

PR #3970, opened 2025-12-10
update(pd): bump builder and base images to v2025.12.7-3-g1c0b8cf-go1.25

purelind:
Bump builder and base images to v2025.12.7-3-g1c0b8cf-go1.25 on master branch.


wuhuizuo:
/hold

we should use the newer image which is building to use go1.25.5.

PR #4075, opened 2026-01-09
refactor(pingcap/tidb): refactor download artifacts steps

wuhuizuo:
- **refactor(pipelines/pingcap/tidb): update TiDB PR check pipeline configuration**
- **feat(pipelines/pingcap/tidb): add utils container and remove debug stage in pipelines**


wuhuizuo:
Some added containers and environment variables is not used in pipelines is not issues, we need it to debugging for migration.

PR #4129, opened 2026-01-29
ticdc: add pulsar heavy integration CI jobs

tenfyzhong:
## Summary
Add TiCDC pulsar heavy integration pipelines and presubmits for latest and next-gen, aligned with heavy resource profiles, and make them optional. 

## Changes
- Add pulsar heavy pipelines/pods and job DSLs for latest and next-gen.
- Add Prow presubmits for pulsar heavy (latest + next-gen include /test pulsar alias, optional).

## Testing
- .ci/update-prow-job-kustomization.sh

### DIFF
--- a/pipelines/pingcap/ticdc/latest/pull_cdc_pulsar_integration_heavy/pod-test.yaml
+++ b/pipelines/pingcap/ticdc/latest/pull_cdc_pulsar_integration_heavy/pod-test.yaml
@@ -9,7 +9,7 @@ spec:
       tty: true
       resources:
         limits:
-          memory: 32Gi
+          memory: 24Gi
           cpu: "6"
   affinity:
     nodeAffinity:


pipelines/pingcap/ticdc/latest/pull_cdc_pulsar_integration_heavy/pod-test.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  securityContext:
+    fsGroup: 1000
+  containers:
+    - name: golang
+      image: "ghcr.io/pingcap-qe/ci/jenkins:v2025.12.28-2-g97bb688-go1.25"
+      tty: true
+      resources:
+        limits:
+          memory: 32Gi
```
wuhuizuo: could we set it to 24GB to keep same with next_gen and lightly job? GKE will scale up with a 64GB memory node since the 32GB memory node can not create the pod.

PR #4143, opened 2026-02-03
pipelines(tiflow): set agent none and simplify pipelines

wuhuizuo:
- Set prow.setPRDescription early. 
- Remove Debug and Checkout stages and drop net-tool/report containers from pod templates. 
- And move checkout into Test stage to address the git folder parsing issue.


pipelines/pingcap/tiflow/latest/ghpr_verify.groovy
```diff
@@ -71,17 +36,21 @@ pipeline {
                     }
                 }
                 stages {
+
```
wuhuizuo:
```suggestion
```

PR #4193, opened 2026-02-27
ci(lib): inject CI annotations by rewriting pod template

dillon-zheng:
Background
yamlFile + yaml overlay does not work reliably in our Jenkins Declarative setup. Even with yamlMergeStrategy, the annotation snippet is ignored, so CI annotations never make it into the final Pod.

What this PR does
Update withCiLabels to read the pod template (via readTrusted), merge CI annotations, and return a full YAML. Pipelines can then use:

yaml pod_label.withCiLabels(POD_TEMPLATE_FILE, REFS)


wuhuizuo (approved):
/hold

need for testing

PR #4311, opened 2026-03-10
ci(tiflash): restructure license check stage in pipelines

wuhuizuo:
Move license-eye download to separate step within utils container and add executable permission before running license check. This improves clarity and ensures proper execution order across all TiFlash pipelines.


wuhuizuo:
tested in: https://do.pingcap.net/jenkins/job/pingcap/job/tiflash/job/pull_integration_test/5339/

PR #4493, opened 2026-04-09
fix(flow): handle hotfix tags on feature release branches

wuhuizuo:
Allow an optional "feature/" prefix in release branch detection and add a
check for hotfix-style prerelease tags (YYYYMMDD-shortcommit). When a feature branch contains such a hotfix tag, return the formatted release version (v + semver) instead of treating it as a regular feature prerelease. This prevents appending feature suffixes to hotfixed releases.


dillon-zheng (approved):
lgtm

PR #4522, opened 2026-04-15
fix(pipeline):  fix job workspaceVolume issue

dillon-zheng:
remove blank volumes line in pod yaml


dillon-zheng:
replay: https://prow.tidb.net/jenkins/job/pingcap/job/tidb/job/ghpr_mysql_test/891/console

PR #4512, opened 2026-04-14
ci: fix workspace pvc rollout for migrated jobs

wuhuizuo:
## Summary
- fix the ineffective workspace PVC rollout from `ci#4509` / `ci#4510`
- move migrated GCP Jenkins jobs to the Jenkins kubernetes plugin's real `workspaceVolume genericEphemeralVolume(...)` field
- remove the dead `workspace-volume` blocks previously added into pod yaml files

## Root cause
The replay log for `pingcap/tidb/release-8.5/pull_build` showed the final created pod still had:
- `workspace-volume: emptyDir`

Even after `pod-pull_build.yaml` defined `spec.volumes[name=workspace-volume].ephemeral.volumeClaimTemplate`.

The Jenkins kubernetes plugin treats workspace storage as a dedicated `workspaceVolume` field. When that field is not set in the declarative agent, it falls back to `EmptyDirWorkspaceVolume`, so a raw pod-yaml volume named `workspace-volume` does not actually replace the default workspace volume.

## Scope
- keep the same rollout scope as the merged phase-1 / phase-2 change
- exclude `PingCAP-QE/ci: seed` as before
- add effective `workspaceVolume genericEphemeralVolume(...)` to the affected Jenkins pipelines
- delete the ineffective `workspace-volume` pod-yaml blocks

## Validation
- confirmed the problem from replay log:
  - https://prow.tidb.net/jenkins/job/pingcap/job/tidb/job/release-8.5/job/pull_build/30
- `111` modified Jenkinsfiles validated successfully via:
  - `https://prow.tidb.net/jenkins/pipeline-model-converter/validate`
- `131` modified pod yaml files parsed successfully with local YAML validation
- rechecked cleanup completeness:
  - no target pod yaml still contains `name: workspace-volume`

PR #4523, opened 2026-04-15
fix(pipeline) fix #4512 blank volumes pod issue

dillon-zheng:
remove blank volumes line in pod yaml


dillon-zheng:
The same as https://github.com/PingCAP-QE/ci/pull/4522

PR #4530, opened 2026-04-16
fix: skip kubectl client/server validation

wuhuizuo:
## Summary
- auto-generate in-cluster kubeconfig from service account token/CA when no kubeconfig context exists in prow job pod
- always run kubectl --dry-run=client and --dry-run=server with strict validation in .ci/verify-k8s-pod-yaml.sh
- update pod-prow_debug pod yaml so this PR triggers pull-verify-k8s-pod-yaml for end-to-end verification
- sync CI guide doc for the new behavior

## Validation
- sh -n .ci/verify-k8s-pod-yaml.sh
- mock kubectl run confirms config set-cluster/set-credentials/set-context and both apply --dry-run=client/server are invoked


dillon-zheng (approved):
lgtm

PR #4564, opened 2026-04-29
fix(tikv/rust-rocksdb): migrate CI from CentOS 7 to Rocky Linux 8

liyishuai:
## Problem

`openssl-sys` v0.9.114 (released ~April 2026) dropped support for OpenSSL < 1.1.0. The current CentOS 7 builder image installs `openssl-devel` from the CentOS 7 repos, which provides OpenSSL **1.0.2k** — now rejected at compile time with:

> "This crate is only compatible with OpenSSL (version 1.1.0, 1.1.1, 3.x, or 4.x), but a different version of OpenSSL was found."

This breaks the `encryption` feature build for **all PRs** targeting `tikv/rust-rocksdb` (confirmed also broken on PR #849, an unrelated change).

Upstream issue: tikv/rust-rocksdb#850

## Fix

Switch all three jobs (`pull-rust-rocksdb-lint`, `pull-rust-rocksdb-test-x86`, `pull-rust-rocksdb-test-arm`) to the same Rocky Linux 8 builder image that `tikv/tikv` latest presubmits already use (`v2026.3.22-3-g892e449`).

Rocky Linux 8 ships OpenSSL 1.1.1 and a modern GCC system-wide, so:
- `yum install -y openssl-devel` is removed (pre-installed in the image)
- The `devtoolset-8` source block is removed (system GCC is sufficient)
- The CentOS-specific `llvm-toolset-7.0` SCL install is replaced with `dnf install -y clang-tools-extra`

## Precedent

`tikv/tikv` latest presubmits already run on this image (see PR #4383).

### DIFF
--- a/prow-jobs/tikv/rust-rocksdb/latest-presubmits.yaml
+++ b/prow-jobs/tikv/rust-rocksdb/latest-presubmits.yaml
@@ -23,7 +23,7 @@ presubmits:
                         - amd64
         containers:
           - name: rust
-            image: ghcr.io/pingcap-qe/cd/builders/tikv:v2025.12.14-1-g33e22ac-centos7-devtoolset8
+            image: ghcr.io/pingcap-qe/cd/builders/tikv:v2026.3.22-3-g892e449
             command: [bash, -ce]
             args:
               - |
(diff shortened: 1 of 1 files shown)


lybcodes:
/lgtm

I reviewed the change and it looks reasonable to me.

The migration from the CentOS 7 builder to `ghcr.io/pingcap-qe/cd/builders/tikv:v2026.3.22-3-g892e449` addresses the OpenSSL compatibility issue for `openssl-sys >= 0.9.114`. The new image supports both `linux/amd64` and `linux/arm64`, and includes Rocky Linux 8.10 with `openssl-devel 1.1.1k`, `gcc/g++ 8.5`, and `cmake`, so removing the old `openssl-devel` install and `devtoolset-8` setup makes sense.

Non-blocking follow-up: before merge please trigger a real `tikv/rust-rocksdb` run for:
- `pull-rust-rocksdb-lint`
- `pull-rust-rocksdb-test-x86`
- `pull-rust-rocksdb-test-arm`
or please trigger them immediately after this config is merged.

wuhuizuo (approved):

PR #4599, opened 2026-05-19
refactor(pipelines): replace hub-zot.pingcap.net/mirrors/hub with GAR mirror

wuhuizuo:
Replace all code-level references of `hub-zot.pingcap.net/mirrors/hub` with `us-docker.pkg.dev/pingcap-testing-account/hub`. Comment-only occurrences (ticdc pipeline comments and tikv/migration doc comments) are left unchanged.

Also simplify the sqllogic test data download loop in `merged_sqllogic_test.groovy` since both sources point to the same location after the replacement.


dillon-zheng (approved):
lgtm

PR #4604, opened 2026-05-20
ci(jobs): add tipipeline shared library to ti-community-infra folder

wuhuizuo:
## Summary
- add the tipipeline folder shared library configuration to 
- copy the same  block used by 
- enable Jenkins jobs under  to load the shared library


wuhuizuo:
Closing as a duplicate of #4603. The diff is identical ( only), and #4603 already has approval plus green checks, so keeping both open would create duplicate merge candidates.

PR #4605, opened 2026-05-20
fix(jobs): set includeInChangesets to false for tipipeline shared library

wuhuizuo:
Fixes FLA-142.\n\nChange `includeInChangesets(true)` to `includeInChangesets(false)` for the tipipeline shared library configured in `ti-community-infra` Jenkins folder, to prevent library changes from triggering unnecessary rebuilds.


wuhuizuo:
Superseded by #4607 which covers all folders, not just ti-community-infra.

PR #4606, opened 2026-05-20
fix(scripts/plugins): match SKIP test status to prevent false .fatal.log, add regression validation

wuhuizuo:
Expand Bazel fatal detection to include SKIP so skip-only test blocks are no longer archived as .fatal.log. Add a regression sample and a shell check covering skip-vs-fail precedence.


dillon-zheng (approved):
lgtm

PR #4610, opened 2026-05-21
chore(libraries): remove pod_label from tipipeline

wuhuizuo:
## Summary
- remove the duplicate pod_label helper from tipipeline
- keep tisys as the single shared implementation
- leave pipeline call sites unchanged

## Testing
- confirmed libraries/tisys/vars/pod_label.groovy still exists
- confirmed the diff only deletes libraries/tipipeline/vars/pod_label.groovy


dillon-zheng (approved):
lgtm

PR #4616, opened 2026-05-21
revert: remove query image tag workflow from ci

wuhuizuo:
This reverts #4615 because the query image workflow must live in tidbcloud/docker-image-controller per the approved design. A replacement workflow PR is being opened there, and the chatops backend defaults are being retargeted in ee-apps.


wuhuizuo (commented):
/approve

Issue #3334, opened 2025-01-24
tikv@release-6.1: Refactor some old pipelines


wuhuizuo:
https://github.com/PingCAP-QE/ci/blob/d3619d0191f4a7bb2fdddcc83079836b09199487/prow-jobs/tikv/tikv/release-6.1-presubmits.yaml#L10-L31

wuhuizuo:
Maintenance Support for TiDB v6.1 ended at 2025-06-13.

Ref: https://www.pingcap.com/tidb-release-support-policy/

PR #4625, opened 2026-05-22
feat(scripts): detect test binary panics in bazel output analyzer

wuhuizuo:
### 1. Analyze panic crash in bazel test output

- Fix section delimiter regex (`=+`) to only match start markers, not pure `====` lines — fixes broken section pairing
- Fix section pairing from `(2p, 2p+1)` to `(p, p+1)` with virtual EOF marker for the last section
- Add `panic:` detection: target output containing `panic:` is suffixed `.panic.log`
- Add `panic:` to `indexReg` so panic lines appear in `bazel-go-test-index.log`
- Add `parse_bazel_go_test_panic_cases` function that writes `crash` entries to `bazel-go-test-problem-cases.json` with:
  - `shard` — which shard crashed
  - `flaky_attempts` — how many times it crashed across retries
  - `cases` — ordered list of test cases in that shard
  - `panic` — panic message

### 2. Deprecate unused scripts

- Move `report-flaky-cases.ts` → `tools/deprecated/plugins/` with symlink at original location
- Move `report_job_result.sh` → `tools/deprecated/plugins/` with symlink at original location

## Testing

Verified against build #3334 of `pull_integration_realcluster_test_next_gen`:
- Old script produced empty `{}` JSON and broken section files
- New script correctly identifies 3 panic crashes in shard 2 and reports them in JSON


wuhuizuo:
All review suggestions addressed in the latest commit:

1. **O(N^2) → O(N)**: panic files processed in a single pass with associative arrays
2. **jq injection fixed**: using `--arg`/`--argjson` instead of string interpolation
3. **`break` → `continue`**: glob loops skip gracefully on missing files
4. **Log validation**: check file exists before `wc -l` for virtual EOF marker
5. **`jq` availability**: `command -v jq` check with early return
6. **Configurable truncation**: `MAX_PANIC_LENGTH` env var (default 200)
7. **Deprecation notices**: added to `report-flaky-cases.ts` and `report_job_result.sh`
8. **Squashed**: all commits into a single clean commit

Please re-review @gemini-code-assist

PR #4661, opened 2026-06-08
ci(ticdc): add legacy safepoint integration tests

wk989898:
This PR introduces ticdc integration tests for the legacy safepoint.
These tests are the same as NEXT_GEN, with the only difference being that LEGACY SAFEPOINT is true.
see https://github.com/pingcap/ticdc/issues/5146


wuhuizuo (approved):

dillon-zheng (approved):
lgtm

PR #4712, opened 2026-06-22
fix(prow): add GCS service account credentials for docs PDF upload

wuhuizuo:
### Problem

rclone in the docs PDF build jobs was failing to upload to GCS with:

```
Error 412: The type of authentication token used for this request
requires that Uniform Bucket Level Access be enabled.
```

Root cause: rclone only set `RCLONE_CONFIG_GCS_TYPE` without specifying a credentials file, causing it to fall back to Application Default Credentials (GCE node SA OAuth token). OAuth tokens require Uniform Bucket Level Access on fine-grained ACL buckets.

### Fix

Add `RCLONE_CONFIG_GCS_SERVICE_ACCOUNT_FILE="/secrets/gcs/service-account.json"` to use the mounted Prow GCS credentials secret, which has proper `storage.objectAdmin` permissions on the bucket.

### Files Changed

- `prow-jobs/pingcap/docs/latest-postsubmits.yaml`
- `prow-jobs/pingcap/docs-cn/latest-postsubmits.yaml`
- `prow-jobs/pingcap/docs-tidb-operator/latest-postsubmits.yaml`

### DIFF
--- a/prow-jobs/pingcap/docs-cn/latest-postsubmits.yaml
+++ b/prow-jobs/pingcap/docs-cn/latest-postsubmits.yaml
@@ -36,6 +36,7 @@ postsubmits:
                 python3 scripts/merge_by_toc.py
                 scripts/generate_pdf.sh
 
+                export RCLONE_CONFIG_GCS_SERVICE_ACCOUNT_FILE="/secrets/gcs/service-account.json"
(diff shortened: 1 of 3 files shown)


wuhuizuo (commented):
/approve

PR #4713, opened 2026-06-22
ci(tekton): remove fetch sync pre-ver image tag trigger

wuhuizuo:
## Summary
- delete `fetch-sync-pre-ver-image-tags-to-hub.yaml`
- remove its `env-prod2` kustomization entry

## Validation
- `kubectl kustomize tekton/v1/triggers/triggers/env-prod2 >/dev/null`
- `bash .ci/update-tekton-kustomizations.sh` (expected to report the intentional tracked diff before commit; no extra file rewrites were needed for this change)


wuhuizuo (commented):
/approve

PR #4714, opened 2026-06-22
ci(tekton): add fake-github triggers to env-gcp

wuhuizuo:
## Summary
- add the `fake-github` trigger set under `tekton/v1/triggers/triggers/env-gcp/_/fake-github`
- register the six new trigger manifests in the `env-gcp` Tekton kustomization
- keep the `env-gcp` wiring aligned with the existing `env-prod2` fake-github trigger implementation

## Validation
- `kubectl kustomize tekton/v1/triggers/triggers/env-gcp >/dev/null`
- `kubectl kustomize tekton/v1/triggers/templates >/dev/null`

## Notes
- `.ci/update-tekton-kustomizations.sh` was run; it reports the checked-in kustomization diff while the new resources are uncommitted, but the resulting resource ordering is aligned.
- `pre-commit` is not installed in this environment, so hook-based checks were not run locally.


wuhuizuo (commented):
/approve

PR #4743, opened 2026-06-24
fix: disable cached checkout for TiFlash pipeline jobs

wuhuizuo:
## Summary
- replace cached checkout calls with direct checkout for TiFlash latest jobs
- apply the same change to TiFlash release-8.5 jobs
- avoid failures when saving oversized Jenkins checkout caches

## Context
The Jenkins checkout cache for this repository has grown to around 11 GB and can fail during cache save. This change temporarily disables cached checkout for the affected TiFlash jobs to prioritize job stability.

## Testing
- not run (pipeline checkout path change only)


dillon-zheng (approved):
lgtm

PR #4724, opened 2026-06-23
fix(scripts): prevent SKIP-only tests from being marked as new_flaky

wuhuizuo:
## Summary

Fixes `parse_bazel_go_test_new_flaky_cases()` in `analyze-go-test-from-bazel-output.sh` so that test cases with only `--- SKIP` evidence (no `--- FAIL`) are **not** classified as `new_flaky`.

## Changes

1. **Expanded grep** (line 93): `"=== RUN|--- PASS"` → `"=== RUN|--- (PASS|FAIL|SKIP)"` — ensures `--- SKIP` and `--- FAIL` lines are included in the candidate count, so skipped tests get count ≥ 2 and are filtered out.

2. **Added `--- FAIL` verification guard** (lines 99–108): Before writing candidates to the result JSON, each is verified to have a `--- FAIL` line in the index. This is defense-in-depth against edge cases in the count logic.

## Testing

- A test case that only shows `=== RUN` + `--- SKIP` (no `--- FAIL`) will **not** be listed in `new_flaky`.
- A test case that shows `=== RUN` + `--- FAIL` in one attempt and `=== RUN` + `--- PASS` in another **is** correctly listed as `new_flaky`.

## Risk

Low — single-file, single-function change. Revert is a single commit.

Fixes: https://github.com/pingcap/tidb/issues/68239

PR #4770, opened 2026-06-30
fix(plugins): fix flaky detection for SKIP-only tests and shard-scoped FAIL verification (FLA-214 PR2)

wuhuizuo:
## Summary

Fix `parse_bazel_go_test_new_flaky_cases()` in `analyze-go-test-from-bazel-output.sh` to correctly handle SKIP-only tests and multi-shard flaky detection.

**Issue**: [FLA-214](mention://issue/fd2c10ca-c3e7-4bcc-8f23-bbc5164bd748)

## Changes

Three changes to `scripts/plugins/analyze-go-test-from-bazel-output.sh`:

1. **Index regex**: Add `SKIP` to `((===|---) (RUN|PASS|FAIL|SKIP))` so SKIP lines are captured in the index file
2. **Flaky detection grep**: Change from `=== RUN|--- PASS` to `=== RUN|--- (PASS|SKIP)` so SKIP-only tests have count=2 (not falsely flagged)
3. **Shard-scoped FAIL verification**: Before adding to `new_flaky`, verify each candidate has `--- FAIL` in the **same shard section** (not globally). Prevents regression from PR #4724 where global FAIL search over-filtered legitimate flaky tests.

## Testing

```
--- TDD tests: flaky detection ---
PASS: TestSplitRangeForTable not in new_flaky (skip_only.log)
PASS: TestFlaky in new_flaky (flaky_two_shards.log)
PASS: TestSkipOnly not in new_flaky (skip_and_flaky.log)
PASS: TestFlaky in new_flaky (skip_and_flaky.log)
All TDD tests passed.
``"

## Risk

Low — targeted changes to a single function. Index regex change affects `parse_bazel_target_output_log` (line 46) but that function uses its own grep on the raw log, not the index file. Revert: single commit revert.

### DIFF
--- a/scripts/plugins/analyze-go-test-from-bazel-output.sh
+++ b/scripts/plugins/analyze-go-test-from-bazel-output.sh
@@ -43,7 +43,7 @@ function parse_bazel_target_output_log() {
                 append="$append.race"
             fi
 
-        elif grep -E "(---|===) (PASS|FAIL|RUN|SKIP)" "$saveFlag.log" | grep -oE "\bTest\w+\b" | sort | uniq -c | grep "^\s*1\b" >/dev/null; then
+        elif grep -E "(---|===) (PASS|FAIL|RUN|SKIP)" "$saveFlag.log" | grep -oE "\bTest[[:alnum:]_/]+\b" | sort | uniq -c | grep "^\s*1\b" >/dev/null; then
             append="$append.fatal"
         fi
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/plugins/analyze-go-test-from-bazel-output.sh [hunk 1]
```diff
@@ -90,12 +90,24 @@ function parse_bazel_go_test_new_flaky_cases() {
             for n in $targetShardOutputLineNum; do
                 local newFlakyCases=($(
                     sed -nE "/^${n}:/,/^[0-9]+:=+ Test output for \//p" "$DEFAULT_GO_TEST_INDEX_FILE" |
-                        grep -E "=== RUN|--- PASS" |
+                        grep -E "=== RUN|--- (PASS|SKIP)" |
                         grep -Eo "\bTest\w+" | sort | uniq -c |
                         grep "^\s*1\b" |
                         grep -Eo "\bTest\w+"
```
wuhuizuo: It is necessary to consider nested test cases, such as TestA/TestB, where the use of `\w+` is inappropriate.

PR #4778, opened 2026-07-01
fix(tekton): rename delivery to deliver in RC2GA task and fix publisher-url defaults

wuhuizuo:
## Summary

- Rename `tag-and-delivery-rc2ga-on-oci-artifacts.yaml` to `tag-and-deliver-rc2ga-on-oci-artifacts.yaml` for consistent naming
- Fix `publisher-url` param in `pingcap-release-ga.yaml`: add `type: string`, remove hardcoded default
- Fix `publisher-url` default in renamed task from `https://publisher.pingcap.net` to `http://publisher.publisher.svc`
- Update kustomization references to match renamed file
- Run `.ci/update-tekton-kustomizations.sh` to sync kustomization resources


tekton/v1/triggers/templates/kustomization.yaml
```diff
@@ -7,6 +7,7 @@ resources:
   - _/build-component-single-platform.yaml
   - _/ci-helper-for-pr.yaml
   - _/collect-multi-arch-image.yaml
+  - _/push-oci-artifact-to-fileserver.yaml
```
wuhuizuo:
```suggestion
```

PR #4830, opened 2026-07-13
fix(prow-jobs/pingcap-qe/ci): remove extra refs and use remote scripts from jsdelivr in periodics

wuhuizuo:


dillon-zheng (approved):
lgtm

PR #4834, opened 2026-07-14
fix(tasks): move step result substitutions to env for Tekton v1 compliance + add presubmit to validate Tekton YAML with server dry-run

wuhuizuo:
## Changes

1. **fix(tasks/zot-validate-repair-image)**: Move cross-step result references (`$(steps.<name>.results.<name>)`) from `script` to `env` — the Tekton v1 admission webhook rejects these in `script` fields.

2. **ci(prow)**: Add `pull-verify-tekton-yaml` presubmit that runs `kubectl apply --dry-run=server` on changed `tekton/v1/**/*.yaml` files, preventing similar admission webhook rejections from reaching main.

## Verification

- Server dry-run of the fixed task YAML passes: ✅
- Original task YAML (before fix) fails with the same admission webhook error: ❌

Closes the issue where `tekton-config-tasks` kustomization was in `Ready=False` state due to webhook rejection.

### DIFF
--- a/prow-jobs/pingcap-qe/ci/presubmits.yaml
+++ b/prow-jobs/pingcap-qe/ci/presubmits.yaml
@@ -117,6 +117,42 @@ presubmits:
                 memory: 256Mi
                 cpu: 200m
 
+    - name: pull-verify-tekton-yaml
+      labels: *labels
+      decorate: true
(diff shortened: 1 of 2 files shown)


wuhuizuo (commented):
/approve

PR #4836, opened 2026-07-15
fix(prow-jobs): switch ee-infra presubmits to OpenTofu and fix private repo clone

wuhuizuo:
Two fixes for ee-infra presubmit jobs:

1. **Private repo clone**: Add `decoration_config.oauth_token_secret` referencing the `github-token` K8s secret. Required because `PingCAP-QE/ee-infra` is a private repo – Prow needs authentication to clone it. This pattern is already used by other private repos like `pingcap-inc/enterprise-plugin`, `pingcap-inc/tici`, etc.

2. **Switch to OpenTofu**: Replace `hashicorp/terraform:1.5.7` with `ghcr.io/opentofu/opentofu:1.6` and use `tofu` CLI instead of `terraform` (both jobs).


wuhuizuo (commented):
/approve

PR #4837, opened 2026-07-15
fix(prow-jobs): add step markers to ee-infra tofu commands

wuhuizuo:
Add `echo "::group::<step-name>"` before each tofu command in ee-infra presubmit jobs so that when a step fails, the log clearly shows which step was running.

Both gcp and tencentcloud jobs are updated.


wuhuizuo (commented):
/approve

PR #4850, opened 2026-07-29
fix(pipelines/tiflash): release-6.5 IT runtime needs mysql client

kumailf:
## Summary
Follow-up to #4849. Rewriting untagged `tiflash-ci-base` to `cd/builders/tiflash:v20231106` unblocked registry pulls, but fullstack tests exec into that container and need `mysql` client:

```text
/bin/sh: mysql: command not found
```

Old `hub.pingcap.net/tiflash/tiflash-ci-base` included mysql (see `jenkins/Dockerfile/release/linux-amd64/tiflash_ci_base`). Builder images do not.

## Change
- Pull `ghcr.io/pingcap-qe/bases/tiflash-base:v1.9.1`
- Locally build `tiflash-ci-base-local:with-mysql` with `yum install -y mysql` (same recipe as historical Dockerfile)
- Use that image for compose rewrites of `tiflash-ci-base` / `tics`

Keeps failpoint fallback from #4849.

## Test plan
- [ ] Merge and re-run `/test pull-integration-test` on pingcap/tiflash#10993
- [ ] Confirm prepare logs show building `tiflash-ci-base-local:with-mysql`
- [ ] Confirm `fullstack-test/sample.test` no longer fails on `mysql: command not found`

PR #4849, opened 2026-07-28
fix(pipelines/tiflash): release-6.5 IT image rewrite + failpoint fallback

kumailf:
## Summary
Unblock `release-6.5` TiFlash `pull_integration_test` (e.g. pingcap/tiflash#10993):

1. **Untagged `tiflash-ci-base` rewrite**  
   `release-6.5` yamls still use `hub.pingcap.net/tiflash/tiflash-ci-base` without a tag. Existing rocky8/rocky9 sed rules miss it, so compose hits the retired registry. Add a sed rule mapping the untagged form to `TIFLASH_TEST_IMAGE` (`ghcr.io/pingcap-qe/cd/builders/tiflash:v20231106`).

2. **TiDB failpoint image fallback** (from closed #4845)  
   If `tidb-server:release-6.5-failpoint` cannot be pulled, fall back to the regular `release-6.5` TiDB image for compose rewrites, and skip `tidb-ci/fail-point-tests` (requires a real failpoint binary).

## Test plan
- [ ] Merge this PR
- [ ] On pingcap/tiflash#10993, run `/test pull-integration-test`
- [ ] Confirm no `hub.pingcap.net` pull for `tics0` / `tiflash-ci-base`
- [ ] If failpoint tag is missing: logs show WARN fallback + skip fail-point-tests; other suites still run
- [ ] If failpoint tag exists: pull succeeds and fail-point-tests still run

PR #4852, opened 2026-07-29
fix(pipelines/tiflash): always skip fail-point-tests on release-6.5

kumailf:
## Summary
Follow-up after #4849 / #4850.

`tidb-ci/cluster.yaml` on release-6.5 uses `qa/tidb:...-failpoint`. The previous logic only skipped `fail-point-tests` when the failpoint image **pull failed**. OCI still has `tidb-server:release-6.5-failpoint`, so pull succeeds, but the image does not expose `enableTestAPI`:

```text
Error: get fail/github.com/pingcap/tidb/server/enableTestAPI
```

## Change
- Always fall back `TIDB_FAILPOINT_IMAGE` → regular `release-6.5` TiDB image
- Always strip `./run-test.sh tidb-ci/fail-point-tests &&` from `run.sh`
- Other suites (`fullstack-test`, async_grpc, collation, …) still run

## Test plan
- [ ] Merge and `/test pull-integration-test` on pingcap/tiflash#10993
- [ ] Logs show WARN skip fail-point-tests
- [ ] `tidb-ci` proceeds to fullstack-test without enableTestAPI error

### DIFF
--- a/pipelines/pingcap/tiflash/release-6.5/pull_integration_test.groovy
+++ b/pipelines/pingcap/tiflash/release-6.5/pull_integration_test.groovy
@@ -267,16 +267,15 @@ pipeline {
                                                         timeout 300 docker pull "${PD_IMAGE}"
                                                         timeout 300 docker pull "${TIKV_IMAGE}"
                                                         timeout 300 docker pull "${TIDB_IMAGE}"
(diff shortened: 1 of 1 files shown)


dillon-zheng (approved):
lgtm

PR #4860, opened 2026-08-05
fix(tekton): add publisher-url param to tidbx-image-distributed trigger

wuhuizuo:
Add missing `publisher-url` parameter to the trigger spec in `update-ops-when-tidbx-image-distributed-to-clouds.yaml`.


wuhuizuo (commented):
/approve

PR #4869, opened 2026-08-10
fix(pipelines/pingcap/tiflash): resource limits in pod-pull_build.yaml

wuhuizuo:
Removed resource requests and limits for memory and CPU.


wuhuizuo (commented):
/approve

PR #4870, opened 2026-08-10
fix(libraries): use codecov bundled in ci image for coverage upload

wuhuizuo:
## Why

`prow.uploadCoverageToCodecov` downloads the codecov binary before every coverage upload, but both download paths are currently broken:

- **GCS mirror** (`storage.googleapis.com/pingcap-ci-bazel-mirror-static-us-central1/...`): the bucket is private (403 without auth), and CI pods have no Workload Identity, so this path is always skipped.
- **Legacy fileserver** (`http://fileserver.pingcap.net/download/cicd/tools/codecov-v0.5.0`): the file has been removed (404) as part of the fileserver sunset; on GCP runners the connection times out.

Result: the `codecov` binary is never downloaded, `chmod` fails, `./codecov` exits with 127, and the `upload coverage to codecov` step fails every run (not transient — retrying the job does not help).

## What changed

In `libraries/tipipeline/vars/prow.groovy` (`uploadCoverageToCodecov`):

1. Prefer the `codecov` binary already bundled in the CI base image (`/usr/local/bin/codecov`, installed from `uploader.codecov.io` in `dockerfiles/ci/base/Dockerfile`).
2. Only when it is missing, download from the public `https://uploader.codecov.io/v0.8.0/linux/codecov` (same source the base image uses, already verified reachable).
3. Remove the GCS mirror and legacy fileserver download paths entirely.

## Verification

- `bash -n` syntax check on the generated script: OK
- `prow.groovy` groovy parse: OK

### DIFF
--- a/libraries/tipipeline/vars/prow.groovy
+++ b/libraries/tipipeline/vars/prow.groovy
@@ -208,36 +208,18 @@ def uploadCoverageToCodecov(refs, flags = "", file = "",  bazelLCov = false, baz
         fi
 
         if [ -f \$coverageFile ]; then
-            # Prefer GCS mirror with Workload Identity token; fallback to legacy fileserver.
-            codecov_url="https://storage.googleapis.com/pingcap-ci-bazel-mirror-static-us-central1/download/cicd/tools/codecov-v0.5.0"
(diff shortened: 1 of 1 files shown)


wuhuizuo (commented):
/approve

PR #4875, opened 2026-08-11
refactor(pipelines): use shared bazel.prepareWorkspace in tidb latest ghpr_check

wuhuizuo:
## What changed

Migrate the first pipeline to the shared bazel workspace helpers added in #4874:

- `pipelines/pingcap/tidb/latest/ghpr_check.groovy`: replace the inline `Hotfix bazel deps URL (temporary)` stage with `bazel.prepareWorkspace()`.

The generated script is byte-identical to the previous inline logic (strip legacy URLs from `WORKSPACE`/`DEPS.bzl` + patch Makefile `check:` target), verified locally against a simulated workspace.

## Why this one first

ghpr_check is a small standalone job — a safe canary to validate the shared helper on a real job before migrating the remaining ~100 pipelines in follow-up PRs.

### DIFF
--- a/pipelines/pingcap/tidb/latest/ghpr_check.groovy
+++ b/pipelines/pingcap/tidb/latest/ghpr_check.groovy
@@ -34,22 +34,10 @@ pipeline {
                 }
             }
         }
-        stage('Hotfix bazel deps URL (temporary)') {
+        stage('Prepare bazel workspace') {
             steps {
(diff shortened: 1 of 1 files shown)


wuhuizuo:
tested in: https://prow.tidb.net/jenkins/job/pingcap/job/tidb/job/ghpr_check/6519/

PR #4876, opened 2026-08-11
refactor(libraries): run bazel workspace prep from shell script

wuhuizuo:
## What changed

Refactor the bazel workspace helpers added in #4874: instead of generating the cleanup script as a Groovy string literal, the logic now lives in a plain shell script.

- **New**: `scripts/pingcap/tidb/prepare-bazel-workspace.sh` — full cleanup logic in plain bash (works with GNU and BSD sed), executed from the ci repo checkout in the agent workspace
- **`bazel.groovy`**: `buildWorkspaceScript`/`buildStaleUrlCleanupScript` replaced by `workspaceEnv()` (pure, computes the `BAZEL_*` env map) + `prepareWorkspace()`/`reapplyStaleUrlCleanup()` which run the script with those env vars
- **`TestBazel.groovy`**: env-map tests + functional tests that actually run the shell script against fixture workspaces (26 cases)

## Why

The previous string-generation approach failed on the Jenkins master (Groovy 2): the `\u005C` unicode escape used to work around a Groovy 5 lexer quirk is an invalid escape there, breaking `prepareWorkspace` at compile time (observed in the ghpr_check replay, Jenkins build 6512). Moving the shell logic into a plain script file removes the escaping problem entirely and makes the logic reviewable/testable as bash.

The public call sites are unchanged (`bazel.prepareWorkspace()`, `bazel.reapplyStaleUrlCleanup()`), so no pipeline migration PR needs changes.

## Verification

- `groovy libraries/tipipeline/tests/TestBazel.groovy`: 26/26 pass on Groovy 5 (local, BSD sed) and on the CI image groovy:4.0.32-jdk17 (GNU sed)
- `workspaceEnv` verified on Groovy 2.4 (Jenkins master runtime)
- `TestComponent` regression: 10/10
- `bash -n` clean

### DIFF
--- a/libraries/tipipeline/tests/TestBazel.groovy
+++ b/libraries/tipipeline/tests/TestBazel.groovy
@@ -1,3 +1,4 @@
+import org.junit.After
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -5,7 +6,8 @@ import org.junit.experimental.runners.Enclosed
 import static org.junit.Assert.*
 
 /**
- * Table-driven unit tests for bazel.groovy utility functions.
+ * Table-driven unit tests for bazel.groovy utility functions and the
+ * resources/bazel/prepare-workspace.sh script.
  *
  * Usage:
  *   groovy libraries/tipipeline/tests/TestBazel.groovy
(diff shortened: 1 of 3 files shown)


wuhuizuo (commented):
/approve

PR #4933, opened 2026-08-20
fix(tekton): make github event simulation deliver events to the event listener

wuhuizuo:
## Summary

The GitHub event simulation script `tekton/tests/test-github-events.ts` failed to trigger any PipelineRun when invoked with kebab-case flags:

```bash
deno run --allow-all tekton/tests/test-github-events.ts   --token=... --url=https://github.com/tikv/pd.git   --ref=refs/heads/release-nextgen-202603   --event-type=push --event-url=http://localhost:8080
```

## Root cause

`parseArgs` from `@std/cli` does not map kebab-case flags (`--event-url`, `--event-type`) to the camelCase fields (`eventUrl`, `eventType`) the script reads. As a result `eventUrl` fell back to the default `https://example.com` and the simulated event was POSTed there instead of to the EventListener, so no PipelineRun was created.

## Changes

- Map kebab-case flags to camelCase fields via `parseArgs` aliases, so both `--eventUrl` and `--event-url` work
- Require `--eventUrl` instead of silently defaulting to `https://example.com`
- Include `created: false` in the push payload so the EventListener group interceptor (`body.created == false`) accepts the event
- Print the HTTP response status and body so future failures are visible immediately

## Verification

- `deno check tekton/tests/test-github-events.ts` passes
- A simulated push (tikv/pd, `refs/heads/release-nextgen-202603`) against the `el-github` EventListener in `ee-cd` successfully created `bp-pd-nextgen-linux-amd64-*` and `bp-pd-nextgen-linux-arm64-*` PipelineRuns

### DIFF
--- a/tekton/tests/test-github-events.ts
+++ b/tekton/tests/test-github-events.ts
@@ -48,6 +48,7 @@ async function generatePushEventPayload(
   return {
     before: "0000000000000000000000000000000000000000",
     after: await getCommitSha(owner, repoName!, ref, client),
+    created: false,
     ref,
     repository: {
       name: repoName!,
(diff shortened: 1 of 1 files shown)


wuhuizuo (commented):
/approve

PR #4936, opened 2026-08-20
refactor(tekton): extract request-kernel-image-sync step into StepAction

wuhuizuo:
## What

- Add a reusable `tekton.dev/v1beta1` `StepAction` `request-kernel-image-sync` under `tekton/v1/step-actions/`
- Update `pingcap-notify-to-deliver-images-to-cloud-tidbx` task to reference the StepAction instead of an inlined step
- Temporarily disable the step (commented out) until the upstream `sync-kernel-images` service is ready

## Why

- StepAction is the reusable unit for a `Step`; the task now composes instead of duplicating the script
- Upstream publisher endpoint `/tidbcloud/sync-kernel-images` is not ready yet, so the step stays disabled

## Notes

- `apiVersion` is `tekton.dev/v1beta1` because StepAction has never been served as `tekton.dev/v1` in any Tekton release; all our clusters run v1.3.1 which serves `v1beta1` (storage)
- StepAction params are passed via `env` since `params` cannot be referenced directly inside StepAction `script`
- Validated with `kubectl apply --dry-run=server` and `kubectl kustomize`

## Validation

- [x] `kubectl apply --dry-run=server` for StepAction and Task
- [x] `kubectl kustomize tekton/v1/step-actions`, `tekton/v1/tasks`
- [x] `.ci/update-tekton-kustomizations.sh`

### DIFF
--- a/tekton/v1/step-actions/kustomization.yaml
+++ b/tekton/v1/step-actions/kustomization.yaml
@@ -1,0 +1,5 @@
+# yaml-language-server: $schema=https://www.schemastore.org/kustomization.json
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - request-kernel-image-sync.yaml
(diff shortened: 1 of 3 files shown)


wuhuizuo (commented):
/approve

PR #4937, opened 2026-08-20
fix(prow-jobs): use serviceAccountName field for ee-infra gcp terraform job

wuhuizuo:
The ee-infra gcp terraform presubmit used `service_account_name` (snake_case) which prow silently drops when parsing the pod spec (the field is `serviceAccountName`), so the job always ran as the unbound default KSA and failed `tofu plan` with 403 IAM_PERMISSION_DENIED.

Switch to the correct `serviceAccountName` field so the job runs as the `terraform-plan-reader` service account, which already has the required project roles (artifactregistry.reader, iam.securityReviewer, etc.).


wuhuizuo (commented):
/approve

PR #4938, opened 2026-08-21
feat(tekton): send kernel image sync images array per stage

wuhuizuo:
## Summary

The publisher `/tidbcloud/sync-kernel-images` endpoint (`request-sync-kernel-image`) changed its payload from a single `image` to an `images` array so multiple images built from the same repo commit are synced in one request (PingCAP-QE/ee-apps#609).

## Changes

`tekton/v1/step-actions/request-kernel-image-sync.yaml`:
- Batch all images from `src` into one request per stage using `{ stage, images: [...] }` instead of one request per image.
- Skip the call when the images list is empty.

## Notes

The upstream endpoint requires all images in one request to share the same repo commit. The step action receives a flat image list via `src`, so it assumes the caller groups images by commit; otherwise the publisher rejects the request.

## Validation

`kubectl kustomize tekton/v1/step-actions` renders cleanly. The yq/jq payload construction was verified locally.

### DIFF
--- a/tekton/v1/step-actions/request-kernel-image-sync.yaml
+++ b/tekton/v1/step-actions/request-kernel-image-sync.yaml
@@ -31,17 +31,20 @@ spec:
 
     printf '%s\n' "$IMAGES_SRC" > /workspace/images.yaml
 
+    images_json=$(yq -o=json '.images // []' /workspace/images.yaml)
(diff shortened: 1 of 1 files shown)


wuhuizuo (commented):
/approve

Issue #4974, opened 2026-08-24
[Seed] Support job DSL seeding for the to Jenkins (master: "0")


wuhuizuo:
## Task: Support job DSL seeding for the **to** Jenkins (`master: "0"`)

The `seed` postsubmit (`prow-jobs/pingcap-qe/ci/postsubmits.yaml`) currently runs on the **from** Jenkins (`labels.master: "1"`) and only processes the job DSL (`jobs/**/*.groovy`) for that Jenkins. As we migrate jobs to the **to** Jenkins, the **to** Jenkins also needs its job definitions kept in sync from the same DSL.

### Goal
Extend the seed mechanism so the job DSL is also processed for the **to** Jenkins (`master: "0"`).

### Scope
- Investigate where the `seed` Jenkins job logic lives (not present as a pipeline in this repo; likely configured on the Jenkins master side).
- Add a mechanism to run the seed against the **to** Jenkins as well (e.g. a second postsubmit targeting `master: "0"`, or extending the seed job itself to seed both instances).
- Keep the existing **from** Jenkins seeding intact.

### Acceptance criteria
- [ ] Job DSL changes are applied to both the **from** and **to** Jenkins.
- [ ] Existing **from** Jenkins seeding behavior unchanged.
- [ ] PR linked to this issue.

Part of #4942

wuhuizuo:
Implementation PR: https://github.com/PingCAP-QE/ci/pull/4977

Approach: convert the `seed` post-submit from a jenkins-agent job to a kubernetes pod with two containers, each triggering the `seed` job on its Jenkins instance via REST (from = `ci-job-jenkins-*`, to = `mig-to-ci-job-jenkins-*`). This avoids the same-name post-submit limitation.

Note for infra side: the **to** Jenkins needs the `seed` Jenkins job available to be triggered.

Issue #4942, opened 2026-08-24
[CI Migration] Migrate jenkins-agent Prow jobs to the target Jenkins (master: "0")


wuhuizuo:
## Goal

Migrate the scheduling of `agent: jenkins` Prow jobs from the current **from** Jenkins to a new **to** Jenkins by flipping the Prow job label `labels.master` from `"1"` to `"0"`.

The **from** Jenkins is **not** deprecated. Only the scheduling of these jobs moves to the **to** Jenkins.

## Mechanism

- Prow routes a ProwJob to a Jenkins controller through the job's `labels` and the controller's `label_selector`:
  - `labels.master: "1"` -> **from** Jenkins (unchanged)
  - `labels.master: "0"` -> **to** Jenkins (after migration)
- The **to** Jenkins is already provisioned with the same pipeline jobs (jobs DSL seeded).
- `*-next-gen` jobs (33) are **excluded** from this migration and stay on `master: "1"` (tracked in Phase 5).

## Scope (non-nextgen jenkins-agent Prow jobs)

Total: **509 jobs** across 17 repos.

| Repo | latest | release | total |
|---|---|---|---|
| pingcap/tidb | 32 | 234 | 266 |
| pingcap/tiflow | 9 | 101 | 110 |
| pingcap-qe/tidb-test | 13 | 15 | 28 |
| pingcap/tiflash | 4 | 20 | 24 |
| pingcap/ticdc | 12 | 7 | 19 |
| tikv/tikv | 2 | 13 | 15 |
| pingcap/tiproxy | 12 | 0 | 12 |
| pingcap-inc/* (tidb/ticdc/tiflow/ee/tiflash-scripts) | 0 | 23 | 23 |
| tikv/pd | 2 | 5 | 7 |
| tikv/migration | 2 | 0 | 2 |
| tikv/copr-test | 1 | 0 | 1 |
| pingcap/tidb-tools | 0 | 1 | 1 |
| PingCAP-QE/ci | 1 | 0 | 1 |

## Phase plan

Tracked through the parent/sub-issue tree below. Each phase issue links its PR(s); a phase issue is closed when its PRs are merged and the post-merge observation on the **to** Jenkins passes.

| Phase | Scope | Issue |
|---|---|---|
| 0 | Verification toolchain (`pull-verify-jenkins-migration` + script + secret wiring) | #4943 |
| 1 | pingcap/tidb `latest` post-submit (10) | #4944 |
[...]

PR #4976, opened 2026-08-24
feat(prow-jobs): add seed post-submit for the to jenkins (master: "0")

wuhuizuo:
Part of #4974 (support job DSL seeding for the **to** Jenkins).

## Summary
The `seed` post-submit currently only seeds the **from** Jenkins (`labels.master: "1"`). Add a `seed-to` post-submit (`labels.master: "0"`, context `seed-to`, same `run_if_changed: ^jobs/.*\.groovy`) so the **to** Jenkins job DSL is processed on the same merges.

## Note
A Prow post-submit `name` maps to the Jenkins job path, so the **to** Jenkins needs a matching `seed-to` Jenkins job (provisioned out-of-band by the infra side) that runs the same Job DSL seeding as `seed` on the **from** Jenkins.

Part of #4942

### DIFF
--- a/prow-jobs/pingcap-qe/ci/postsubmits.yaml
+++ b/prow-jobs/pingcap-qe/ci/postsubmits.yaml
@@ -4,6 +4,11 @@ global_definitions:
     labels:
       master: "1"
     decorate: false # need add this.
+  jenkins_job_to: &jenkins_job_to
+    agent: jenkins
+    labels:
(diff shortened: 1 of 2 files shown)


wuhuizuo:
superseded: seed becomes a kubernetes pod that triggers the seed job on both jenkins instances (same-name postsubmits are not allowed)

Issue #4954, opened 2026-08-24
Phase 3: tikv/tikv latest (2)


wuhuizuo:
### Jobs
- `tikv/tikv/pull_integration_test`
- `tikv/tikv/pull_unit_test`

### Steps
- [ ] Flip `labels.master` `"1"` -> `"0"` for these jobs.
- [ ] `/test pull-verify-jenkins-migration`; all pass on **to** Jenkins.
- [ ] Merge PR; observe. Close this issue.

### PRs
- TODO: PR link

wuhuizuo:
Closing this phase issue: both `tikv/tikv` `latest` jenkins-agent jobs (`pull_integration_test`, `pull_unit_test`) now have `labels.master: "0"` on `main` (verified at 12285887).

Migrated via #4990.

Issue #4946, opened 2026-08-24
Phase 3: Migrate latest jobs of other repos


wuhuizuo:
## Phase 3 — Migrate `latest` jobs of other repos

**Status: ✅ DONE** — all 10 children are closed.

One issue per repo, one PR per issue (repo `latest` only). Ordered roughly by impact.

### Children

- [x] tiflow `latest` (9) — #4949
- [x] pingcap-qe/tidb-test `latest` (13) — #4950
- [x] pingcap/ticdc `latest` (12) — #4951 (closed 2026-09-16)
- [x] pingcap/tiproxy `latest` (12) — #4952
- [x] pingcap/tiflash `latest` (4) — #4953
- [x] tikv/tikv `latest` (2) — #4954
- [x] tikv/pd `latest` (2) — #4955
- [x] tikv/migration `latest` (2) — #4956
- [x] tikv/copr-test `latest` (1) — #4957
- [x] PingCAP-QE/ci `seed` post-submit (1) — #4958

### Steps

- [x] Each child: flip `labels.master` `"1"` -> `"0"` for the repo's latest jobs, verify via `/test pull-verify-jenkins-migration`, merge, observe.
- [x] Close children as they complete; close this parent when all children are closed.

wuhuizuo:
Progress: 9 of 10 child phase issues are now closed. Only #4951 (`pingcap/ticdc` `latest`, 5 jobs remaining) is still open, so this issue stays open.

wuhuizuo:
✅ Done — all 10 repo children are closed; every non-nextgen `latest` job of these repos now routes to the target Jenkins.

PR #4990, opened 2026-08-25
chore(prow-jobs): migrate tikv/tikv latest jobs to target jenkins

wuhuizuo:
Part of #4954 (Phase 3: tikv/tikv latest).

## Summary
Flip `labels.master` `"1"` -> `"0"` for the 2 jenkins-agent `latest` jobs in `prow-jobs/tikv/tikv/latest-presubmits.yaml` so they are scheduled on the **to** Jenkins.

## Verification
Hold before merge; run `/test pull-verify-jenkins-migration` and observe on the **to** Jenkins after merge.

Part of #4946
Part of #4942


wuhuizuo:
**Analysis (autonomous run)** — verification failed on 1 job + 1 infra-fail.

- `tikv/tikv/pull_unit_test`: **infra-fail** (executor/checkout contention on the single online executor).
- `tikv/tikv/pull_integration_test`: `failed` — needs investigation (could be version mismatch or real test failure); the to-Jenkins run should be checked against the from-Jenkins baseline result for the same params.
- No success to split.

Suggestion: retry when capacity allows; if pull_integration_test fails again, compare against from-Jenkins result before deciding.

PR #4995, opened 2026-08-25
fix(tekton): correct stage param reference in request-kernel-image-sync

wuhuizuo:
## Problem

`flux get kustomizations -A` shows the `ee-cd` kustomizations are not ready:

- `tekton-config-step-actions`: StepAction `request-kernel-image-sync` dry-run failed - non-existent variable in `$(params.stage)`: spec.env[TARGET_STAGES]
- `tekton-config-tasks` and `tekton-config-pipelines` fail as dependencies

## Root cause

Introduced in #4940: the StepAction `request-kernel-image-sync` referenced `$(params.stage)`, but the param is defined as `stages`.

## Fix

- tekton/v1/step-actions/request-kernel-image-sync.yaml: `$(params.stage)` -> `$(params.stages)`

## Test

Ran `flux get kustomizations -A --status-selector ready=false` after sync to confirm the kustomizations become ready.

### DIFF
--- a/tekton/v1/step-actions/request-kernel-image-sync.yaml
+++ b/tekton/v1/step-actions/request-kernel-image-sync.yaml
@@ -29,7 +29,7 @@ spec:
     - name: PUBLISHER_URL
       value: $(params.publisher-url)
     - name: TARGET_STAGES
-      value: $(params.stage)
+      value: $(params.stages)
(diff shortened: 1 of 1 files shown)


wuhuizuo (commented):
/approve

Issue #5010, opened 2026-08-26
[Toolchain] Apply matrixCache to the Checkout & Prepare stage to save resources


wuhuizuo:
### Background
The tidb `ghpr_check2` pipeline (and similar tidb pipelines) have a `Checkout & Prepare` stage that runs `prow.checkoutRefsWithCacheLock`, builds `tidb-server`, downloads PD/TiKV components, and prepares the bazel workspace, then `stash name: 'ws'` for the matrix test pods.

Builds #7/#8 of `ghpr_check2` re-ran this expensive prepare because the earlier build was interrupted during checkout/prepare. `matrixCache` is currently only applied to the **Test** stage (skip already-passed matrix cells). We want to also skip `Checkout & Prepare` when a previous build for the same refs already prepared successfully.

### Feasibility notes
- `matrixCache.shouldSkip/markDone` is keyed by (org/repo/baseSha/pullShas/stageName/extraParams); it can be applied to any stage (no matrix axis -> fixed key).
- Caveat 1: `markDone` only fires on success; build#7->#8 both failed at prepare, so cache would not have skipped them. Benefit applies only to later re-runs after a successful prepare.
- Caveat 2: the `ws` stash is per-build. If `Checkout & Prepare` is skipped on a re-run, the matrix `unstash 'ws'` has nothing to restore. The prepared workspace must be recoverable independently (persistent stash/cache keyed by refs) OR the skip must rely on the existing git/binary/bazel caches to make re-prepare cheap.
- Existing caches (git objects, tidb-server binary, bazel) already reduce re-prepare cost; the matrixCache adds an explicit skip.

### Acceptance criteria
- [ ] `Checkout & Prepare` can be skipped on re-run when a prior build for the same refs succeeded.
- [ ] The matrix test stage still gets the prepared workspace (no `unstash` failure).
- [ ] Existing `Test` stage matrixCache behavior unchanged.
- [ ] PR linked to this issue.

Part of #4942

wuhuizuo:
Implementation PR: https://github.com/PingCAP-QE/ci/pull/5011

Approach (safe): guard the expensive prepare sub-steps (make server, pd/tikv download) behind `matrixCache.shouldSkip(REFS,'Prepare',[:])` / `markDone`; the binary/components caches restore the binaries on re-runs, and `checkoutRefs`, `bazel.prepareWorkspace`, `mv`/`touch`, and `stash 'ws'` always run so the matrix `Test` stages always get a complete workspace.

[...]

PR #5011, opened 2026-08-26
feat(pipelines): skip ghpr_check2 prepare sub-steps on matrix cache hit

wuhuizuo:
Part of #5010 (Apply matrixCache to the Checkout & Prepare stage).

## Summary
Apply the existing `matrixCache` to the expensive sub-steps of the `Checkout & Prepare` stage in `pipelines/pingcap/tidb/latest/ghpr_check2.groovy`, so re-runs of the same refs skip the costly work when a previous build already prepared successfully.

## Approach (safe: the ws stash always happens)
- `prow.checkoutRefsWithCacheLock` always runs.
- `make server` is guarded by `matrixCache.shouldSkip(REFS, 'Prepare', [:])` (rebuild also when `bin/tidb-server` is missing, in case the binary cache was evicted).
- PD/TiKV download is guarded by the same key inside a `cache(path: "./bin", includes: 'pd-server,tikv-server', key: prow.getCacheKey('components', REFS))` block, so re-runs restore them from cache while skipping the network download.
- `matrixCache.markDone(REFS, 'Prepare', [:])` is written only after tidb-server build + component download succeed.
- `bazel.prepareWorkspace()`, the `mv`/`touch`, and `stash name: 'ws'` always run -> the matrix `Test` stages always get a complete workspace (the user requirement).
- Existing `Test` matrixCache usage is unchanged.

## Notes
- `matrixCache` lives in `libraries/tisys/vars/matrixCache.groovy` (loaded via the tipipeline library); the key is `MD5(org/repo/baseSha/pullShas/stageName/extras)`.
- Only `latest/ghpr_check2.groovy` shares the exact structure; the `release-*` branches have byte-identical copies and could be a follow-up.

Part of #4942

### DIFF
--- a/pipelines/pingcap/tidb/latest/ghpr_check2.groovy
+++ b/pipelines/pingcap/tidb/latest/ghpr_check2.groovy
@@ -37,16 +37,27 @@ pipeline {
                     script {
                         prow.checkoutRefsWithCacheLock(REFS, timeout = 5, credentialsId = GIT_CREDENTIALS_ID)
                     }
+                    // Skip the expensive prepare sub-steps on re-runs of the same refs; the
+                    // cache restores keep the workspace complete for the stash below.
                     cache(path: "./bin", includes: 'tidb-server', key: prow.getCacheKey('binary', REFS)) {
-                        sh label: 'tidb-server', script: 'ls bin/tidb-server || make server'
+                        script {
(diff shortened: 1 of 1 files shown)


wuhuizuo:
Superseded: the Checkout & Prepare sub-steps already have caches (git/binary/components/bazel), so an explicit matrixCache guard is unnecessary. Keeping the repo unchanged.

PR #4987, opened 2026-08-25
chore(prow-jobs): migrate pingcap/ticdc latest jobs to target jenkins

wuhuizuo:
Part of #4951 (Phase 3: pingcap/ticdc latest).

## Summary
Flip `labels.master` `"1"` -> `"0"` for the 25 jenkins-agent `latest` jobs in `prow-jobs/pingcap/ticdc/latest-presubmits.yaml` so they are scheduled on the **to** Jenkins (pull_cdc_* integration tests and ghpr_* checks).

## Verification
Hold before merge; run `/test pull-verify-jenkins-migration` and observe on the **to** Jenkins after merge.

Part of #4946
Part of #4942

PR #5037, opened 2026-08-28
chore(prow-jobs): mark pull-verify-jenkins-migration as optional

wuhuizuo:
Align `pull-verify-jenkins-migration` with the design in #4942 (optional, not a tide gate). It was missing `optional: true`, so tide treats it as required and blocks migration PRs whose verification can't pass due to target-Jenkins infra/version issues (e.g. #4987).


wuhuizuo:
Reverting — do NOT make pull-verify-jenkins-migration optional. Will force-merge #4987 instead.

PR #5066, opened 2026-09-01
feat(libraries): add Git HTTP askpass helper

wuhuizuo:
## Summary
- add a small Git `GIT_ASKPASS` helper for Jenkins HTTP authentication
- return the configured username or password only for the matching Git prompt
- fail closed for unsupported prompts

## Dependency
This is the first PR in a two-PR fix for Git CDN authentication. The companion `prow.groovy` integration PR is based on this branch and must be merged after this PR.

## Validation
- exercised username and password prompts with test environment variables
- verified unsupported prompts return a non-zero exit code
- `git diff --check`

### DIFF
--- a/libraries/tipipeline/scripts/git_askpass.sh
+++ b/libraries/tipipeline/scripts/git_askpass.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+case "${1-}" in
+Username*)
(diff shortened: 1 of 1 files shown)


wuhuizuo (commented):
/approve

PR #5063, opened 2026-08-31
feat(libraries): support Git CDN HTTP credentials

wuhuizuo:
## Summary
- keep the existing `credentialsId` behavior, including SSH checkout for private repositories
- use `GIT_SSH_CREDENTIALS_ID` as a default checkout-scope SSH credential for jobs with private submodules
- use `GIT_HTTP_CREDENTIALS_ID` and a temporary `GIT_ASKPASS` helper when `GIT_CDN_ENABLED=true`
- let Git try anonymous CDN access first and provide the PAT only after an HTTP 401
- keep the askpass helper outside the workspace so `git clean -ffdx` and pipeline cache do not remove or store it
- skip the unnecessary GitHub host-key scan when GitHub SSH URLs are being rewritten by the CDN

## Behavior

- GCP: existing direct SSH/HTTPS behavior remains unchanged when `GIT_CDN_ENABLED` is false.
- TKE: both HTTPS and SSH-style GitHub URLs can be rewritten by Jenkins global Git configuration; private requests use the HTTP credential after a 401.
- `refs/pull/*` refspecs and recursive submodule checkout remain unchanged.

## Dependency

The companion `PingCAP-QE/ee-ops` PR adds the matching Jenkins credentials and TKE Git rewrite environment. Merge this PR before enabling the feature in ee-ops.

## Validation

- `groovy -e 'new GroovyShell().parse(new File("libraries/tipipeline/vars/prow.groovy"))'`
- Git `insteadOf` verification for HTTPS, `git@github.com:` and `ssh://git@github.com/` URL forms
- `git diff --check`

PR #5067, opened 2026-09-01
feat(libraries): integrate Git HTTP askpass

wuhuizuo:
## Summary
- route GitHub HTTPS and SSH-style URLs through the existing Git CDN configuration when enabled
- inject Jenkins' HTTP credential only through a temporary `GIT_ASKPASS` scope
- preserve direct SSH authentication for GCP and private submodules
- skip GitHub host-key scanning only when the URL is rewritten to the in-cluster CDN
- clean up the askpass helper even when checkout fails

## Why the original CI missed this
The existing `pull-test-jenkins-libraries` job only executed `libraries/tipipeline/tests/Test*.groovy`. It did not load every Jenkins shared-library entrypoint, so a syntax error in `vars/prow.groovy` was only discovered when a production Jenkins job loaded the library.

This PR adds `TestProw.groovy`, which is intentionally a direct Groovy loadability check. Unlike the existing JUnit-style tests, a parser exception propagates as a non-zero process exit code in the current Prow command.

## Test-first verification
- `589c8934` intentionally reproduced the missing-brace syntax error from #5063.
- The remote [`pull-test-jenkins-libraries` check failed](https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/PingCAP-QE_ci/5067/pull-test-jenkins-libraries/2094724859945816064) at `prow.groovy:73`.
- `25430e3a` restored the missing closure brace.
- The same remote check passed after the fix: [`pull-test-jenkins-libraries`](https://prow.tidb.net/view/gs/prow-tidb-logs/pr-logs/pull/PingCAP-QE_ci/5067/pull-test-jenkins-libraries/2094725612122935296).

## Dependency
This is the second PR in the two-PR fix. It depends on [#5066](https://github.com/PingCAP-QE/ci/pull/5066), which provides the askpass helper used by this PR. #5066 must be merged first; this PR is now based on `main` so its CI can validate the complete change. After #5066 merges, the current two-file diff remains the effective PR2 change.

## Validation
- `groovy libraries/tipipeline/tests/TestProw.groovy`
- all existing Jenkins shared library tests
- `git diff --check`

## Behavior
- GCP with `GIT_CDN_ENABLED=false`: existing direct SSH/HTTPS behavior remains unchanged.
- TKE with `GIT_CDN_ENABLED=true`: GitHub HTTPS and SSH-style URLs can be rewritten to the in-cluster HTTP CDN; public repositories remain anonymous, while private requests receive Jenkins' PAT through `GIT_ASKPASS` after an HTTP authentication challenge.


wuhuizuo:
Addressed the askpass cleanup review:

- Initialize `tmpAskPassScript` before the `try` block.
- Keep temporary file creation and askpass resource loading inside `try`.
- Only run cleanup when the temporary path is non-empty.
- Preserve the original `mktemp` failure instead of issuing a misleading cleanup command.
- Added a JUnit 4 regression test covering temporary askpass creation failure.

Validation: `groovy libraries/tipipeline/tests/TestProw.groovy` (7 tests, 0 failures).

wuhuizuo (commented):
/approve

PR #5068, opened 2026-09-01
fix(libraries): place askpass helper in resources

wuhuizuo:
## Summary

- move the Git askpass helper into the Jenkins Shared Library `resources/` directory
- keep the existing `libraryResource 'scripts/git_askpass.sh'` call unchanged
- add a JUnit regression test that verifies the resource is present at the runtime lookup path

## Root cause

After #5067 was merged, TKE Jenkins build 11 reached the Git checkout stage but failed with:

```
ERROR: No such library resource scripts/git_askpass.sh could be found.
```

The helper was stored at `libraries/tipipeline/scripts/git_askpass.sh`. Jenkins `libraryResource` only loads files from the shared library's `resources/` directory, so the runtime path must be:

```
libraries/tipipeline/resources/scripts/git_askpass.sh
```

The failure was reproduced by [TKE Jenkins staging TiDB ghpr_build #11](https://do.pingcap.net/jenkins-staging/job/pingcap/job/tidb/job/ghpr_build/11/).

## Validation

- `groovy libraries/tipipeline/tests/TestProw.groovy` (8 tests, 0 failures)
- `sh -n libraries/tipipeline/resources/scripts/git_askpass.sh`
- `git diff --check`

Follow-up to [#5067](https://github.com/PingCAP-QE/ci/pull/5067).


wuhuizuo (commented):
/approve

PR #5072, opened 2026-09-02
fix(libraries): create git askpass helper in shared workspace

wuhuizuo:
## Summary

Fix empty Git askpass helper that broke private submodule clones (e.g. tiflash `merged_build` cloning `tidbcloud/cloud-storage-engine`) when Git URLs are rewritten to the in-cluster git-cdn.

## Root cause

`withGitAskPass` created the helper via `mktemp` under `/tmp`, then populated it with Jenkins `writeFile`:

- `sh` steps run in the pod tool container (as root), so `/tmp/git-askpass.XXXXXX` lived in that container's filesystem.
- `writeFile` executes through the agent FilePath in the **JNLP** container, a different filesystem.
- Result: the file Git executed (`GIT_ASKPASS`) was the empty file left by `mktemp`; askpass returned no credentials; GitHub answered `Repository not found / Authentication failed`.

## Fix

- Put the helper under the shared workspace volume instead of `/tmp` — the volume is mounted in both the agent and the tool container.
- Keep it outside the Git worktree (every caller clones into a workspace subdirectory via `dir()`, so the workspace root is never touched by `git clean`/checkout).
- Let `writeFile` create the file so its owner is the agent user; a file pre-created by an `sh` (`mktemp`) step would be owned by the container's root user and unwritable by the agent (AccessDenied).
- Unique name via UUID to avoid collisions.

Verified: with a faithful agent pod (same builder image, git 2.43.5, real creds, same env), the private submodule clones successfully through git-cdn; an empty askpass reproduces the original error byte-for-byte.

### DIFF
--- a/libraries/tipipeline/vars/prow.groovy
+++ b/libraries/tipipeline/vars/prow.groovy
@@ -63,8 +63,15 @@ def withGitAskPass(String credentialsId, Closure body) {
     def tmpAskPassScript
 
     try {
-        tmpAskPassScript = sh(script: 'mktemp "${TMPDIR:-/tmp}/git-askpass.XXXXXX"', returnStdout: true).trim()
+        // Keep the askpass script inside the shared workspace volume but outside
+        // the Git worktree. Every caller clones into a workspace subdirectory via
+        // dir(), so the workspace root is never touched by git clean/checkout.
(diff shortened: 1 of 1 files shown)


wuhuizuo (commented):
/approve

Issue #4967, opened 2026-08-24
Phase 4: pingcap/tidb-tools (1)


wuhuizuo:
### Jobs
- `pingcap/tidb-tools/pull_verify` (file: `prow-jobs/pingcap/tidb-tools/presubmits.yaml`)

### Steps
- [ ] Flip `labels.master` `"1"` -> `"0"`, `/test pull-verify-jenkins-migration`, merge, observe.
- [ ] Close this issue when complete.

### PRs
- TODO: PR link

wuhuizuo:
pingcap/tidb-tools pull_verify migrated via #5100. Merged (job is independently red on both Jenkins due to #5101 sunset-fileserver leftover, unrelated to migration).

Issue #4947, opened 2026-08-24
Phase 4: Migrate release-branch jobs


wuhuizuo:
## Phase 4 — Migrate release-branch jobs

**Status: in progress — 8 of 9 children closed; 21 jobs remain deferred.**

### Children

| Child | Scope | State |
|---|---|---|
| #4959 | pingcap/tidb release branches | ✅ closed (117 jobs) |
| #4960 | pingcap/tiflow release branches | ✅ closed |
| #4961 | pingcap-qe/tidb-test release branches | ✅ closed |
| #4962 | pingcap/tiflash release branches | ✅ closed |
| #4963 | pingcap/ticdc release branches | ✅ closed |
| #4964 | tikv/tikv release branches | ✅ closed (10 jobs, 2026-09-16) |
| #4965 | tikv/pd release branches | ✅ closed |
| #4966 | pingcap-inc repos | ✅ closed (2 jobs; the 21 "heavy" jobs below were explicitly deferred) |
| #4967 | pingcap/tidb-tools | ✅ closed |

### Remaining (deferred)

21 `agent: jenkins` presubmits still carry `labels.master: "1"` on `main` (`41edbfd8`). They were deferred when #4966 was closed, and are tracked here:

| Repo | Branch | Jobs |
|---|---|---|
| pingcap-inc/ticdc | release | 7 |
| pingcap-inc/tidb | release-8.5 | 8 |
| pingcap-inc/tiflow | release-8.5 | 4 |
| pingcap-inc/tiflash-scripts | all | 2 |

- `pingcap-inc/tiflash-scripts` (2 jobs, both `optional: true`) is covered by #5250, in flight.
- The other 19 (`pingcap-inc/ticdc` 7, `pingcap-inc/tidb` 8, `pingcap-inc/tiflow` 4) have no PR yet; one PR per repo/branch is the intended shape.

### Steps
- [ ] All branches migrated and observed.
- [ ] Close this issue when complete.

wuhuizuo:
All Phase 4 release-branch children closed: tidb(#4959) tiflow(#4960) tidb-test(#4961) tiflash(#4962) ticdc(#4963) tikv(#4964) pd(#4965) pingcap-inc(#4966) tidb-tools(#4967). Phase 4 complete.

wuhuizuo:
## Phase 4 migration progress

更新时间：2026-09-09 02:22 UTC

[...]

PR #5100, opened 2026-09-04
ci(prow): migrate pingcap/tidb-tools pull_verify to target jenkins

wuhuizuo:
Part of #4967 (Phase 4: pingcap/tidb-tools).

## Summary
Flip `labels.master` `"1"` -> `"0"` for the jenkins-agent presubmit job `pingcap/tidb-tools/pull_verify` in `prow-jobs/pingcap/tidb-tools/presubmits.yaml` so it is scheduled on the **to** Jenkins.

## Note
This job has build history on the from Jenkins (last builds FAILURE), so `/test pull-verify-jenkins-migration` can copy real parameters.

Part of #4947
Part of #4942

### DIFF
--- a/prow-jobs/pingcap/tidb-tools/presubmits.yaml
+++ b/prow-jobs/pingcap/tidb-tools/presubmits.yaml
@@ -4,7 +4,7 @@ presubmits:
     - name: pingcap/tidb-tools/pull_verify
       agent: jenkins
       labels:
-        master: "1"
+        master: "0"
       decorate: false # need add this.
       always_run: true
       skip_report: false


wuhuizuo:
Merging despite `pull-verify-jenkins-migration` = fail.

**Verification result — migration is sound, job is independently broken (tracked in #5101):**
- The verify triggered `pull_verify` on the to Jenkins with real parameters copied from the from-Jenkins `lastBuild` (build #2). Build #1 on the to Jenkins ran the full pipeline (~15 min: checkout, build linux/arm64, unit tests) and failed at the same Integration Test step as the from Jenkins.
- **Identical root cause on both Jenkins:** the pipeline downloads the `dumpling` artifact from the decommissioned `http://sunset-fileserver.pingcap.net`, whose DNS no longer resolves (`curl: (6) Could not resolve host: sunset-fileserver.pingcap.net`). The from-Jenkins builds #1/#2 fail for this same reason — this is a pre-existing pipeline leftover (see #5101), unrelated to the `master: "1" -> "0"` flip.
- Conclusion: behavior on the to Jenkins is equivalent to the from Jenkins; the migration config change is correct.

Part of #4967 / #4947 / #4942.

PR #5105, opened 2026-09-04
fix(.ci): replay inline pod yaml as top-level var, not inline expression

wuhuizuo:
## Problem
`.ci/replay-jenkins-build.sh` inlines a changed pod yaml into the replayed pipeline as:
```groovy
yaml new String(java.util.Base64.decoder.decode("..."), 'UTF-8')
```
The declarative pipeline sandbox rejects method calls evaluated **inside the agent `yaml` field** — every replay using pod-yaml inline failed at script start with `groovy.lang.MissingPropertyException: No such property: UTF` (observed on enterprise-extensions pr-verify and tidb-tools pull_verify replays).

## Fix
Decode the pod yaml into a **top-level `final` variable declared before `pipeline {`** and reference the variable in the agent (`yaml _REPLAY_POD_<VAR>`). The assignment runs in the normal script context, so the sandbox accepts it.

## Validation
With this pattern, tidb-tools pull_verify replay #5 reached **SUCCESS** (OCI-download integration test ran end-to-end); EE pr-verify replay progressed past script start to real pod scheduling.

### DIFF
--- a/.ci/replay-jenkins-build.sh
+++ b/.ci/replay-jenkins-build.sh
@@ -186,7 +186,9 @@ build_inline_script_with_pod_yaml() {
 
     local found=0
     : > "$out_file"
-    local line out_line b64 prefix comment repl
+    local -a out_lines=()
+    local -a preludes=()
+    local line out_line b64 prefix comment repl pvar
     while IFS= read -r line; do
         out_line="$line"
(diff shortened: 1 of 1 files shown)


wuhuizuo (commented):
/approve

Issue #4964, opened 2026-08-24
Phase 4: tikv/tikv release branches (13)


wuhuizuo:
## Phase 4 — tikv/tikv release branches (13 jobs)

**Status: ✅ DONE** — every job in every existing branch file carries `labels.master: "0"` on `main` (`41edbfd8`).

The branch table below originally listed 9 branches / 13 jobs, but only **6** branch files exist in this repository (`release-8.2`, `release-8.3` and `release-8.4` have no presubmit file), so the actual scope is **10 jobs**.

### Branches
| Branch | Jobs | `master: "0"` |
|---|---|---|
| release-6.5 | 2 | 2 |
| release-7.1 | 2 | 2 |
| release-7.5 | 2 | 2 |
| release-8.1 | 2 | 2 |
| release-8.5 | 1 | 1 |
| release-9.0-beta | 1 | 1 |

### Steps
- [x] All branches migrated and observed.
- [x] Close this issue when complete.

### PRs
- `release-8.1` — #5156 (merged 2026-09-16)
- earlier branches — #5125–#5130, #5153–#5158, #5166

### Notes
- `release-8.1/pull_unit_test` has a pre-existing flaky suite. #5156 was merged with `--admin` against a red `pull-verify-jenkins-migration`, using a manually triggered green target-side build (`pull_unit_test` `#15`, same `JOB_SPEC`/revision/pod template) as evidence — the same approach used for #5157. Details are in the #5156 discussion.
- `release-8.1/pull_integration_test` is `optional: true` and its verify step stays **skipped**, because the from-Jenkins has never built that job (`nextBuildNumber: 1`), so there is no `JOB_SPEC` to copy.

wuhuizuo:
tikv/tikv release branches migrated via #5125-#5130. Done via Phase 4 heavy batch.

wuhuizuo:
Status check (verified against `main` @ 12285887): 3 of the 13 `tikv/tikv` release-branch jobs are still on `master: "1"`:

- `tikv/tikv/release-8.1/pull_integration_test` and `tikv/tikv/release-8.1/pull_unit_test` — PR #5156 (open)
- `tikv/tikv/release-8.5/pull_unit_test` — PR #5157 (open)

Keeping this issue open until #5156/#5157 land and pass verification.

wuhuizuo:
[...]

PR #25, opened 2020-12-15
Oncall refine rules

lilinghai:
The failed job
[tikv_ghpr_integration_common_test-29125](https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/tikv_ghpr_integration_common_test/detail/tikv_ghpr_integration_common_test/29125/pipeline/)

Issue #10, opened 2020-11-15
remove 'dead' containers of ci k8s cluster


shuke987:
https://dashboard.pingcap.net/dashboard/3

PR #5156, opened 2026-09-04
ci(prow): migrate tikv/tikv release-8.1 jobs to target jenkins

wuhuizuo:
Part of #4964.

## Summary
Flip `labels.master` `"1"` -> `"0"` for the jenkins-agent jobs in `prow-jobs/tikv/tikv/release-8.1-presubmits.yaml` so they are scheduled on the **to** Jenkins.

Part of #4947 / #4942

### DIFF
--- a/prow-jobs/tikv/tikv/release-8.1-presubmits.yaml
+++ b/prow-jobs/tikv/tikv/release-8.1-presubmits.yaml
@@ -10,7 +10,7 @@ presubmits:
     - name: tikv/tikv/release-8.1/pull_unit_test
       agent: jenkins
       labels:
-        master: "1"
+        master: "0"
       decorate: false # need add this.
       skip_if_only_changed: *skip_if_only_changed
       context: pull-unit-test
@@ -21,7 +21,7 @@ presubmits:
     - name: tikv/tikv/release-8.1/pull_integration_test
       agent: jenkins
       labels:
-        master: "1"
+        master: "0"
       decorate: false # need add this.
       always_run: false  # update here after test passed
       optional: true # update here after test passed


wuhuizuo:
**Merging with `pull-verify-jenkins-migration` red — evidence summary**

This PR only flips `labels.master` from `"1"` to `"0"` for the two `tikv/tikv/release-8.1` jobs; every other check on this PR is green. The red verify is a **pre-existing shared flake** in the `pull_unit_test` job, not a migration regression.

### What the verify actually reports

- `pull_integration_test`: **skipped** — the from-Jenkins job `tikv/tikv/release-8.1/pull_integration_test` has **zero builds** (`nextBuildNumber: 1`), so verify has no `JOB_SPEC` to copy. The job is `optional: true`, so it does not block. Unrelated to this PR.
- `pull_unit_test`: FAILURE on the target side, caused by flaky tests that also fail on the from side.

### Root cause of the dominant failure

`engine_traits util::tests::bench_sequence_number_window` (`components/engine_traits/src/util.rs:213`) is racy by construction:

- `produce_random_seqno(16, 100000)` spawns **16 producer threads** that allocate global sequence numbers from a shared `AtomicU64`, then collects them **in `mpsc::Receiver::recv` order** — which is not the sequence-number allocation order.
- The benchmark loop then pushes that same out-of-order sequence into `SequenceNumberWindow` repeatedly and asserts `committed_seqno <= sequence.number`.
- The observed failure is exactly that inversion: `committed_seqno 3311, seqno3310`.

Hard failures observed (all 3 retries exhausted):

| build (target) | `bench_sequence_number_window` |
| --- | --- |
| `#16` / `#17` / `#18` | FAIL 3/3 — three consecutive builds |
| `#15` | FAIL, FAIL, PASS |
| from `#25` / `#43` | PASS |

The second recurring failure, `sst_importer import_mode::tests::test_import_mode_switcher` / `_timeout` ABRT with `pthread lock: Invalid argument`, is a RocksDB teardown race (`pthread_mutex_lock` returning `EINVAL` means use-after-destroy; see facebook/rocksdb#11349 and rust-rocksdb#720). It cannot be a resource problem: `pthread_mutex_lock` does not return `EINVAL` on resource exhaustion (that would be `EAGAIN`/`ENOMEM`), and the test's own assertions pass — the process aborts during teardown.

### Why the target side fails more often

| side | recent pass rate |
| --- | --- |
| target (`do.pingcap.net/jenkins-staging`) `#10`–`#18` | **1 / 9** SUCCESS |
| from (`prow.tidb.net/jenkins`) `#24`–`#43` | 9 SUCCESS / 9 FAILURE / 2 ABORTED |

All the failing tests are timing/scheduling sensitive, and both sides run the **same pod template from this repository**, so a pod-spec difference cannot explain the asymmetry.

Resources were explicitly tested and ruled out. An A/B replay of the same `JOB_SPEC` on the target with `limits.cpu: 8` (current) vs `limits.cpu: 16` gave SUCCESS (`#15`) and FAILURE (`#16`, on a different test) respectively — raising the CPU limit does not help. Note `#5227` already raised this job to 8 cpu / 32Gi.

### Merge evidence

Target-side build `#15` succeeded with the same `JOB_SPEC` (`rev-38a3c48-69f2772_e7150e7`), the same pod template and the same revision — both partitions fully green (2023 + 1982 tests, 0 hard failures). This is the same class of evidence used to merge `#5157` on 2026-09-15 (a manually triggered target-side build standing in for a red verify).

### Follow-ups (do not block this migration)

1. File a `tikv/tikv` issue for the racy `bench_sequence_number_window` (16 producers + receive-order collection + monotonicity assertion).
2. File a `tikv/tikv` issue for the `import_mode::tests::*` RocksDB teardown race.
3. Quarantine both in the `[profile.ci]` nextest profile — the `retries = 2` in `tikv/tikv`'s `.config/nextest.toml` cannot help a test that fails 3/3 — and/or improve `pull-verify-jenkins-migration` to compare the same revision on the from side before classifying a build as `failed`.

PR #5136, opened 2026-09-04
ci(prow): migrate pingcap/tidb release-7.5 jobs to target jenkins

wuhuizuo:
Part of #4959.

## Summary
Flip `labels.master` `"1"` -> `"0"` for the jenkins-agent jobs in `prow-jobs/pingcap/tidb/release-7.5-presubmits.yaml` so they are scheduled on the **to** Jenkins.

Part of #4947 / #4942

PR #5163, opened 2026-09-04
ci(prow): migrate pingcap/tidb release-7.5 verified jobs to target jenkins

wuhuizuo:
Split from #5136 — only the jobs that verified **SUCCESS** on the to Jenkins (verify runid 2095811292349599744):
- `ghpr_check`
- `ghpr_unit_test`
- `pull_e2e_test`


wuhuizuo:

PR #5164, opened 2026-09-04
ci(prow): migrate pingcap/tidb release-8.1 verified jobs to target jenkins

wuhuizuo:
Split from #5137 — only the jobs that verified **SUCCESS** on the to Jenkins (verify runid 2095811327330095104):
- `ghpr_check`
- `ghpr_check2`
- `ghpr_unit_test`
- `pull_br_integration_test`
- `pull_e2e_test`


wuhuizuo:

PR #5165, opened 2026-09-04
ci(prow): migrate pingcap/tiflow release-7.5 verified jobs to target jenkins

wuhuizuo:
Split from #5142 — only the jobs that verified **SUCCESS** on the to Jenkins (verify runid 2095811541558366208):
- `ghpr_verify`
- `pull_cdc_integration_pulsar_test`
- `pull_dm_compatibility_test`
- `pull_dm_integration_test`


wuhuizuo:

PR #5166, opened 2026-09-04
ci(prow): migrate tikv/tikv release-7.5 verified jobs to target jenkins

wuhuizuo:
Split from #5155 — only the jobs that verified **SUCCESS** on the to Jenkins (verify runid 2095812213846577152):
- `pull_unit_test`


wuhuizuo:

PR #5180, opened 2026-09-08
fix(artifacts): discard incomplete OCI downloads

wuhuizuo:
## Problem

The JDBC integration build from [PingCAP-QE/ci#5178](https://github.com/PingCAP-QE/ci/pull/5178) received a truncated TiKV archive after an OCI HTTP/2 stream failure. `tar` partially created `tikv-server`; the Jenkins retry then treated that partial file as a completed download and it crashed with SIGSEGV.

## Fix

- write OCI blobs to a temporary file and publish them only after a successful fetch
- extract archives into a temporary directory before publishing the requested path
- remove remaining archives and partial extraction paths before retrying

## Validation

- `bash -n scripts/artifacts/download_pingcap_oci_artifact.sh`
- local mocked archive test: a corrupt first archive removes both partial output and archive; the next invocation downloads and extracts successfully
- `git diff --check`

`pre-commit` is not installed in the local environment.

### DIFF
--- a/scripts/artifacts/download_pingcap_oci_artifact.sh
+++ b/scripts/artifacts/download_pingcap_oci_artifact.sh
@@ -5,17 +5,36 @@ set -eo pipefail
 function fetch_file_from_oci_artifact() {
     local oci_url="$1"
     local to_match_file="$2"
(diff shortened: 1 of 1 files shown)


wuhuizuo (commented):
/approve

PR #5184, opened 2026-09-08
refactor(pipelines): reorganize tiflash pull_unit_next_gen job layout

wuhuizuo:
Reorganize the pingcap/tiflash `pull_unit_next_gen` job layout to align with `pull_unit_test`:

- Move `pull_unit_next_gen.groovy` into `pull_unit_next_gen/pipeline.groovy` and update the job DSL script path accordingly.
- Add a dedicated `pull_unit_next_gen/pod-test.yaml` pod template instead of relying on the previous (non-existent) top-level `pod-pull_unit-test.yaml` reference.
- Derive the pod template file path from `JOB_BASE_NAME` in both `pull_unit_next_gen` and `pull_unit_test` pipelines.
- Cache ccache under `$WORKSPACE/.ccache` (workspace volume) and drop the shared `customWorkspace` (tiflash-build-common).

### DIFF
--- a/jobs/pingcap/tiflash/latest/pull_unit_next_gen.groovy
+++ b/jobs/pingcap/tiflash/latest/pull_unit_next_gen.groovy
@@ -14,7 +14,7 @@ pipelineJob('pingcap/tiflash/pull_unit_next_gen') {
     definition {
         cpsScm {
             lightweight(true)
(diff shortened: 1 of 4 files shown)


wuhuizuo (commented):
/approve

PR #5191, opened 2026-09-09
fix(pipelines): redirect ghpr DDL tmp with postStart

wuhuizuo:
## Problem

The previous postStart symlink implementation did not solve the failure. Replay build 28 confirms the hook ran, the workspace mount was declared writable, and the temporary target was assigned to UID/GID 1000; TiDB still received `read-only file system` under `make bazel_*`.

## Change

Keep the postStart ownership setup, but replace the symlink with a bind mount:

1. create `/home/jenkins/agent/tidb-tmp` on the workspace PVC and set `1000:1000` / `0775`;
2. remove the image-precreated `/tmp/tidb` before tests begin;
3. bind-mount the temporary directory at `/tmp/tidb`, then explicitly remount that bind mount read-write;
4. invoke the existing Bazel ConfigMap script through `/bin/sh`.

The privileged `golang` container already has the capability required for bind mounts. A separate mount point prevents the Bazel test sandbox from resolving a symlink target that it makes read-only.

## Validation

- `git diff --check`
- Kubernetes pod template client schema validation
- Ruby YAML parse
- Extracted postStart shell syntax check
- staging replay: build 29 (source build 25)

### DIFF
--- a/pipelines/pingcap/tidb/latest/ghpr_check2.groovy
+++ b/pipelines/pingcap/tidb/latest/ghpr_check2.groovy
@@ -89,7 +89,7 @@ pipeline {
                         )
                     }
                 }
-                agent{
+                agent {
                     kubernetes {
                         namespace K8S_NAMESPACE
                         defaultContainer 'golang'
@@ -103,7 +103,7 @@ pipeline {
                     expression { return !matrixCache.shouldSkip(REFS, 'Test', [script_and_args: env.SCRIPT_AND_ARGS]) }
                 }
                 stages {
-                    stage('Test')  {
+                    stage('Test') {
                         environment {
                             CODECOV_TOKEN = credentials('codecov-token-tidb')
                         }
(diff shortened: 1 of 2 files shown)


wuhuizuo:
## Jenkins Replay Status

- [ ] `pipelines/pingcap/tidb/latest/ghpr_check2.groovy` + `pipelines/pingcap/tidb/latest/pod-ghpr_check2.yaml` → `pingcap/tidb/ghpr_check2`
  - [replay build 28](https://do.pingcap.net/jenkins-staging/job/pingcap/job/tidb/job/ghpr_check2/28/) confirmed the ownership-corrected symlink hook reached the pod, but TiDB still saw a read-only path under Bazel; it was stopped after diagnosis.
  - [replay build 29](https://do.pingcap.net/jenkins-staging/job/pingcap/job/tidb/job/ghpr_check2/29/) is building from [build 25](https://do.pingcap.net/jenkins-staging/job/pingcap/job/tidb/job/ghpr_check2/25/). It uses the bind-mount postStart implementation, with the mount explicitly remounted read-write.
  - inline pod YAML is enabled.

PR #5197, opened 2026-09-10
chore(pipelines): use ghcr.io release image instead of ghcr-remote mirror

wuhuizuo:
## Summary

Replace the `us-docker.pkg.dev/pingcap-testing-account/ghcr-remote/pingcap-qe/cd/utils/release` image reference with the upstream `ghcr.io/pingcap-qe/cd/utils/release` in all pod YAMLs under `pipelines/`.

## Changes

- 89 pod YAML files under `pipelines/` updated.
- Image tags are kept unchanged; only the registry/repository prefix is switched from the ghcr-remote mirror to ghcr.io.


wuhuizuo (commented):
/approve

PR #5228, opened 2026-09-14
fix(tidb): wait for PD leader and TiKV stores in real-tikv tests

wuhuizuo:
## Problem

The `ghpr_check2` real-tikv matrix branches are flaky. In build `ghpr_check2 #229` (`pingcap/tidb#71029`) the `bazel_addindextest1` and `bazel_importintotest3` branches failed, and the rest were aborted by `parallelsAlwaysFailFast()`.

The archived TiKV logs show all/some TiKV processes crashing during startup:

```
[FATAL] [lib.rs:563] ["failed to get timestamp from PD: Other(\"[components/pd_client/src/tso.rs:102]: Timestamp channel is dropped\")"]
  <server::server::TikvServer<...>>::init  at components/server/src/server.rs:434:55
```

With no TiKV store registered, PD never bootstraps the cluster, so every test in the branch fails at `CreateMockStoreAndSetup` with:

```
type:NOT_BOOTSTRAPPED message:"cluster is not bootstrapped"
```

## Root cause

`run_real_tikv_tests.sh` started TiKV right after `wait_for_pd_cluster` saw 3 healthy members on `/pd/api/v1/health`. That endpoint only reports member health; it does **not** guarantee the PD leader (and its TSO service) is ready.

Timeline from `bazel_addindextest1`:

```
08:19:43.032 TiKV first TSO request -> "requested pd is not leader of cluster"
08:19:43.033 PD leader "ready to serve"
08:19:46.390 TiKV FATAL "Timestamp channel is dropped"
```

TiKV's `TikvServer::init` fetches a timestamp while starting and panics if that request races the leader election. It is a per-branch race (~a few ms window), which is why it hits a couple of branches per run and the job fails most of the time.

## Fix

- Wait for the PD leader to be elected (`/pd/api/v1/leader`) before starting TiKV.
- Add a short grace period so the leader finishes bringing up the TSO service.
- Replace the fixed `sleep 10` with a wait until all 3 TiKV stores are `Up` (`/pd/api/v1/stores`). This also turns a crashed TiKV into an early, explicit failure instead of a long run of `cluster is not bootstrapped` test failures.

## Notes

- Not related to the PR under test; TiKV is a `master` artifact downloaded by the pipeline and crashes before any TiDB interaction.
- `integrationtest_with_tikv.sh` has the same "health then start TiKV" pattern and can get the same fix in a follow-up.
- Verified with `bash -n` and `shellcheck` (only pre-existing warnings remain).


wuhuizuo:
## Jenkins Replay Status

Replaying the failing build `ghpr_check2 #229` with the PR version of `scripts/pingcap/tidb/run_real_tikv_tests.sh` injected into the checked-out ci repo.

**Result: PASS** ✅ — the startup race that caused `cluster is not bootstrapped` is gone.

| # | replay build | result | note |
|---|---|---|---|
| 1 | https://do.pingcap.net/jenkins/job/pingcap/job/tidb/job/ghpr_check2/237/ | FAILURE | TiKV crash gone. Failed on unrelated flaky tests: `bazel_importintotest4` (`TestImportInto/TestGlobalSortBasic`: `[executor:8173]PreCheck failed: there is active job on the target table already`) and `bazel_txntest` (aborted by `parallelsAlwaysFailFast()`). |
| 2 | https://do.pingcap.net/jenkins/job/pingcap/job/tidb/job/ghpr_check2/243/ | SUCCESS | Re-ran the two branches that failed in attempt 1 (matrix cache skipped the already-passed ones): `bazel_importintotest4` 311/311 pass, `bazel_txntest` 57/57 pass. |

Fix verification (across both attempts):
- `[FATAL] ... Timestamp channel is dropped`: **0** occurrences
- `cluster is not bootstrapped`: **0** occurrences
- New waits worked in every executed branch: `PD leader is ready` -> `TiKV cluster is ready`
- Failed branches' TiKV logs (attempt 1) show no FATAL; TiKV ran until teardown.

Summary: `success=2` (attempt 2 passed), `failure=1` (attempt 1, unrelated flaky test), `skipped=0`. The attempt-1 failure was an unrelated test-isolation flake, not the PD/TiKV startup race.

D3Hunter (approved):

PR #26, opened 2020-12-17
WeCom report and tweaks.

KivenChen:

PR #23, opened 2020-12-11
add: new rule

shuke987:
add a new rule.

PR #5245, opened 2026-09-16
chore(pipelines/ticdc): raise resources for ticdc light integration test pods

wuhuizuo:
## What

Raise the `golang` container resources of the three `pingcap/ticdc` light
integration-test pod templates to 24Gi / 8 cpu:

- `pipelines/pingcap/ticdc/latest/pull_cdc_storage_integration_light/pod-test.yaml`
  (16Gi / 6 cpu -> 24Gi / 8 cpu)
- `pipelines/pingcap/ticdc/latest/pull_cdc_mysql_integration_light/pod-test.yaml`
  (16Gi / 4 cpu -> 24Gi / 8 cpu)
- `pipelines/pingcap/ticdc/latest/pull_cdc_kafka_integration_light/pod-test.yaml`
  (16Gi / 6 cpu -> 24Gi / 8 cpu)

This matches the memory of the heavy templates and gives the light jobs the
same headroom, while staying well inside the node capacity.

## Why

These three jobs have been failing on the target staging Jenkins
(`do.pingcap.net/jenkins-staging`) since 2026-09-07/08 with a staging-only
mechanism: PD's embedded etcd stalls for several seconds (`kv gets too slow`,
`keep alive lease too slow`, `clock offset` with a multi-second jet lag on a
50ms refresh interval), the PD leader lease expires, PD steps down, and the CDC
servers all exit with `ErrCaptureSuicide`, so the test CLI fails. The same
commit passes on the from-Jenkins (GKE). The stalled paths are the PD/TiKV data
directories under `/tmp`, which the pods only get 6 (mysql light: 4) cpu to
serve while PD + TiDB + TiFlash + 3x TiKV + 3x CDC + the downstream services all
run inside the same cgroup.

Evidence: four replays of the same JOB_SPEC (`pingcap/ticdc#6207 @55270ded`) on
staging, with the pod template inlined into the replay script so that only the
resource limits differ:

| replay | pod resources | result | PD stall |
| --- | --- | --- | --- |
| `#23` | 16Gi / 6 cpu | FAILURE | yes |
| `#24` | 24Gi / 8 cpu | SUCCESS (9/16 groups) | no |
| `#25` | 24Gi / 8 cpu | SUCCESS (16/16 groups) | no |
| `#26` | 16Gi / 6 cpu | FAILURE | yes |

The concurrent load was comparable in the passing and the failing windows
(16-17 concurrent jobs on the production Jenkins instance, which shares the same
TKE cluster), so the earlier failures are not explained by a quiet window.

This unblocks the migration presubmits `#5219`, `#5218` and `#5234`, which are
currently red only because of this environment failure.

## Validation

- `yq` parses all three templates.
- `.ci/verify-k8s-pod-yaml.sh` passes (`kubectl` client and server dry-run).
- `.ci/verify-storage-class-policy.sh` passes.
- The replay builds above used exactly these limits (`requests=limits` = 8 cpu /
  24Gi) and ran the full 16-group matrix green.


wuhuizuo (commented):
/approve

PR #5250, opened 2026-09-16
docs(prow-jobs): note why pingcap-inc/tiflash-scripts jobs stay on the from Jenkins

wuhuizuo:
## What

Keep `labels.master: "1"` for the two `pingcap-inc/tiflash-scripts`
jenkins-agent jobs and add a short in-code note explaining why they cannot be
migrated to the to Jenkins yet:

- `pingcap-inc/tiflash-scripts/pull_regression_test`
- `pingcap-inc/tiflash-scripts/pull_schrodinger_test`

The `labels.master` flip from the earlier revision of this PR is reverted, so
this is now a **comment-only change with no functional effect**.

Part of #4942.

## Why

Both jobs are blocked on the to Jenkins: the pipeline has to `yum install`
runtime packages inside the CentOS 7 based pod image, and the package source it
is pinned to (`vault.centos.org`) answers HTTP 403 from the to Jenkins cluster
while it still answers 200 from the from Jenkins cluster. The
`Prepare Runtime And Artifacts` stage therefore fails before any test case runs,
on every build.

Since these jobs work on the from Jenkins today, flipping them now would make
them permanently red, so they stay on the from Jenkins until the package source
problem is solved.

Evidence (replays, build logs and in-cluster probes from both clusters): see the
comment below.

## Verification

- `.ci/update-prow-job-kustomization.sh` (no change), YAML parse, and
  `git diff --check` all pass.
- `git diff origin/main` for the touched file is comment-only.


wuhuizuo:
### Why these two jobs cannot be migrated yet (evidence)

`pull-verify-jenkins-migration` replayed both jobs on the to Jenkins (staging). Both fail in the **`Prepare Runtime And Artifacts`** stage, and the test stages are skipped — so **no test case ever runs**:

| job | to Jenkins build | result / duration | failed stage |
| --- | --- | --- | --- |
| `pull_regression_test` | https://do.pingcap.net/jenkins-staging/job/pingcap-inc/job/tiflash-scripts/job/pull_regression_test/1/ | FAILURE / 7 min | `Prepare Runtime And Artifacts` |
| `pull_schrodinger_test` | https://do.pingcap.net/jenkins-staging/job/pingcap-inc/job/tiflash-scripts/job/pull_schrodinger_test/1/ | FAILURE / 22 min | same |

The failing step is the `yum install` in `pipelines/pingcap-inc/tiflash-scripts/latest/common/prepare_runtime_and_artifacts.sh` (the image does not ship these packages, so every pod installs them at runtime):

```
yum install -y jq java-1.8.0-openjdk-headless fio mariadb lsof net-tools nmap-ncat
http://vault.centos.org/centos/7/os/x86_64/repodata/repomd.xml: [Errno 14] HTTP Error 403 - Forbidden
failure: repodata/repomd.xml from base: [Errno 256] No more mirrors to try.
ERROR: script returned exit code 1
```

#### Not a pipeline difference

Both sides run the **same** pipeline and the **same** image:

- the from Jenkins jobs also fetch `pipelines/pingcap-inc/tiflash-scripts/latest/...` from `PingCAP-QE/ci@main`;
- pod image on both sides: `ghcr.io/pingcap-qe/cd/builders/tiflash:v2025.8.10-2-gc9e3144-centos7`;
- identical `yum install` command, identical list of missing packages.

The from Jenkins ran the very same thing successfully in April 2026 (build `#23` / `#24`, both SUCCESS), so the difference is not in the configuration.

#### A/B on the same JOB_SPEC

Replayed `pingcap-inc/tiflash-scripts` PR 2003 (`@5cf7c340`, base `52ab023e`) on both sides:

| side | build | `yum install` | what happened next |
| --- | --- | --- | --- |
| from (`prow.tidb.net/jenkins`) | https://prow.tidb.net/jenkins/job/pingcap-inc/job/tiflash-scripts/job/pull_regression_test/24/ | **OK** — 47 packages, `Total download size: 57 M`, `Complete!` | entered `Run Regression Test`, cases passing |
| to (`do.pingcap.net/jenkins-staging`) | `#1` | **403** | prepare stage exited, test stage skipped |

#### In-cluster probes, same image, from inside both clusters

| URL | from cluster (GKE) | to cluster (TKE) |
| --- | --- | --- |
| `http://vault.centos.org/centos/7/os/x86_64/repodata/repomd.xml` | 301 (redirect to https) | **403** |
| `https://vault.centos.org/centos/7/os/x86_64/repodata/repomd.xml` | 200 | **403** |
| `http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=stock` | DNS NXDOMAIN | DNS NXDOMAIN |
| `https://mirrors.cloud.tencent.com/centos-vault/7.9.2009/os/x86_64/repodata/repomd.xml` | 200 | 200 |
| `https://mirrors.aliyun.com/centos-vault/7.9.2009/os/x86_64/repodata/repomd.xml` | 200 | 200 |
| `https://mirrors.tuna.tsinghua.edu.cn/centos-vault/7.9.2009/os/x86_64/repodata/repomd.xml` | 200 | 200 |
| `https://mirrors.fedoraproject.org/mirrorlist?repo=epel-7&arch=x86_64` | 200 | 200 |

Probed with `kubectl exec` into running pods that use this exact image: the from Jenkins pod of build `#24` on the GKE cluster, and production pods on the TKE cluster.

So after CentOS 7 EOL the mirror list host is gone and `vault.centos.org` serves the from cluster but answers **HTTP 403 to the to cluster's egress**, which makes the runtime package install impossible there.

#### The image's repo config (why the from side still works)

`/etc/yum.repos.d/CentOS-Base.repo` inside that image has the mirror list commented out and hard-codes the vault host:

```ini
[base]
#mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=os&infra=$infra
baseurl=http://vault.centos.org/centos/$releasever/os/$basearch/
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
# [updates] / [extras] same shape; [centosplus] enabled=0
```

#### What would unblock the migration

1. Make the package source reachable from the to cluster (e.g. point the CentOS 7 repos at a live vault mirror, or bake the packages into the image). Note that the mirror paths differ from the vault layout — `.../centos-vault/7.9.2009/...` works, `.../centos-vault/7/...` returns 404 — so it is not a plain host swap.
2. Or move these jobs to a Rocky 8 based builder image (as `pingcap/tiflash` itself uses), which needs checking against what the regression/schrodinger scripts expect.

Until one of those is done, these two jobs stay on the from Jenkins. This PR only documents the blocker in the job definitions.

#### Unrelated: `pull-verify-prow-jobs-run`

That optional check is red for a different reason: it resolves the default-branch head SHA with an anonymous `git ls-remote` against `pingcap-inc/tiflash-scripts`, which is a private repository, so it cannot resolve it. It is not caused by this change and does not block merging.

PR #5279, opened 2026-09-22
fix(artifacts): prefer inline OCI config data over fetching the config blob

wuhuizuo:
## Summary

`scripts/artifacts/download_pingcap_oci_artifact.sh` prints the artifact config for debugging via `oras manifest fetch-config`. That command does not honor the OCI 1.1 inline `data` field and always fetches the config blob.

Third-party artifacts (etcd, minio, ycsb, schema-registry, ...) use the `application/vnd.oci.empty.v1+json` config, whose content is embedded in the manifest as `data: "e30="` (the base64 of `{}`). When the registry cannot serve that blob, the fetch fails and `set -eo pipefail` aborts the entire download — even though the manifest is intact and the actual layers download fine.

Observed on `pingcap/tiflow/pull_cdc_integration_storage_test`:

```
+ download_pingcap_oci_artifact.sh ... '--etcdctl=v3.5.15' ...
🚀 start download etcdctl
📦 == artifact information =======
Error: GET "https://cr-qcloud.pingcap.net/v2/mirrors/hub/pingcap/third-party/etcd/blobs/sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a": mismatch Content-Length
ERROR: script returned exit code 1
```

`sha256:44136fa3...` is the OCI empty config; the mirror stores it as 0 bytes, while the manifest already carries the content inline.

## Changes

- Add `fetch_artifact_config()`:
  - if the manifest's `.config.data` is present, print it decoded (`yq -r '.config.data | @base64d'`);
  - otherwise fall back to `oras manifest fetch-config`.
- In `download()`, call it and tolerate a failure (`|| echo "⚠️ failed to fetch artifact config (ignored)"`) so the informational step can never break the download.

## Test plan

- [x] etcd (inline `data`): prints `{}`, then the layer is fetched and `etcdctl` is extracted.
- [x] tidb (real config, no inline `data`): falls back to `oras manifest fetch-config` and still prints the annotations.
- [x] Verified against `cr.pingcap.net/mirrors/hub` (the same zot deployment as `cr-qcloud.pingcap.net`).

## Notes

- Registry-side, the empty config blob is genuinely 0 bytes on the zot mirror; this change only stops CI from depending on it. Clients that fetch the blob directly are still affected.


wuhuizuo (commented):
/approve

PR #5283, opened 2026-09-22
refactor(jobs): harden the one-folder-per-job migration (no symlinks)

wuhuizuo:
## Summary

This is the hardening change that must land **before** the remaining migration
batches. It fixes a crash in the migration tool, drops the back-compat symlink
strategy, and repoints the CI consumers and Prow triggers that still only looked
at `pipelines/**`.

## Why no symlinks

The original design kept a relative symlink at every legacy path so that old
`scriptPath` values and the seed job kept working. That is unsafe here: the seed
runs with

```groovy
removedJobAction('DELETE')
removedConfigFilesAction('DELETE')
```

so any job the seed fails to discover is **deleted**. A missing or dangling
symlink therefore escalates into a deleted Jenkins job (and, for a folder
definition, into a deleted folder subtree). A symlink-based back-compat layer
also behaves differently across git checkout, glob and Jenkins Job DSL target
resolution, which is hard to validate exhaustively.

Instead:

- **The seed scans both trees** (`PingCAP-QE/ee-ops#2413`):
  `targets('jobs/**/*.groovy\njenkins/jobs/**/*.groovy')`. Discovery no longer
  depends on migration progress, and the legacy pattern becomes a no-op once the
  migration finishes.
- **The migration moves the DSL** (so a job is never defined twice) and **copies**
  the pipeline and pod templates into the job folder. The legacy `pipelines/`
  copies stay until `--cleanup`, which keeps the previous `scriptPath` values
  resolvable until the seed re-indexes — the protection the symlinks provided,
  without symlinks.
- **The checker fails when a job is defined in both trees**, so the two layouts
  can never both own a job.

## Bug fix

A job without any pod template left `sed_args` empty, so
`sed -E "${sed_args[@]}" <Jenkinsfile>` passed the file path as the sed *script*
and aborted the run (`sed: ...: invalid command code f`). Under `set -euo
pipefail` the tool stopped mid-job: the DSL had been moved, the back-compat
symlink had not been created, and a re-run could not see the job any more.

Reproduced against the real pre-migration pilot slice, where `prow_debug` is a
no-pod job:

```
$ scripts/migrate-jenkins-jobs.sh --root <slice> --apply
...
sed: 1: "/var/folders/...": invalid command code f
EXIT CODE = 1   migrated folders: 1   jobs/ back-compat symlinks: 0
```

The fixture now contains a no-pod job, and the migration test asserts it is
migrated.

## Consumer and trigger repointing

Migrated jobs were silently skipped by several checks because their
`run_if_changed` / path filters only matched `pipelines/**`:

| Consumer | Change |
|---|---|
| `pull-replay-jenkins-pipelines` | also run for `jenkins/jobs/**/Jenkinsfile` |
| `pull-verify-jenkins-pipelines` | also run for `jenkins/jobs/**/Jenkinsfile` |
| `pull-verify-k8s-pod-yaml` | also run for `jenkins/jobs/**/pod*.yaml` |
| `pull-verify-secret-scan` | include `jenkins/**` in `run_if_changed` |
| `seed` postsubmit | also trigger for `jenkins/jobs/**/*.groovy` |
| `.ci/verify-jenkins-credential-policy.sh` | scan `jenkins/**` changes |
| `.ci/replay-jenkins-build.sh` | map `jenkins/jobs/<org>/<repo>/<branch>/<job>/Jenkinsfile` to the same Jenkins job path as the legacy layout; resolve the pod next to the Jenkinsfile; include it in `--auto-changed` |
| `.github/renovate.json` | also track pod YAML under `jenkins/jobs` |
| `.agents` replay skill | document the new layout |

## Testing

```
bash .ci/test-migrate-jenkins-jobs.sh              # 53 checks
bash .ci/test-check-jenkins-job-references.sh      #  9 cases
bash .ci/test-verify-jenkins-credential-policy.sh  # regression cases
```

All pass. `bash -n` is clean for every touched script and `git diff --check`
reports no whitespace issues. `verify-jenkins-pipelines.sh` needs a Jenkins
instance and is exercised by the presubmit, not locally.

## Not in this PR

The pilot's back-compat symlinks (`jobs/ti-community-infra/...`) are **not**
removed here. Removing them before the seed scans `jenkins/jobs/**` would delete
those jobs, so that is a separate follow-up PR to merge **after**
`PingCAP-QE/ee-ops#2413` is deployed.

## Merge order

1. Deploy `PingCAP-QE/ee-ops#2413` (seed scans both trees).
2. Merge this PR.
3. Rebase the migration stack (`#5257` … `#5275`) onto this branch and drop the
   `jobs/**` symlink additions from each layer.
4. `--cleanup` last, once every job is migrated and a staging replay passed.


wuhuizuo:
## Stack rebased on this change

The migration stack (`#5257` … `#5275`) has been rebased onto this branch and regenerated with the de-symlinked tool, so every layer now:

- moves the job DSL to `jenkins/jobs/<...>/<job>/dsl.groovy`,
- copies the pipeline and pod templates next to it,
- creates **no** `jobs/**` symlinks (verified: zero `120000` entries added by any layer),
- leaves `pipelines/**` untouched (retired by `--cleanup`, not by the migration),
- passes `.ci/check-jenkins-job-references.sh`.

Totals after the full stack: 374 `dsl.groovy`, 374 `Jenkinsfile`, 444 `pod*.yaml`, 65 `aa_folder.groovy`, and the checker reports `374 job(s) and 818 reference(s)`, all resolving.

The 19 layers are all drafts for now, so nothing merges out of order.

## Merge order

1. `PingCAP-QE/ee-ops#2413` — seed scans `jobs/**` **and** `jenkins/jobs/**`. Required before any migration layer: each layer deletes `jobs/**/<job>.groovy`, and the seed runs with `removedJobAction('DELETE')`.
2. **This PR** (`#5283`) — safe to merge on its own; it only hardens tooling and repoints consumers/triggers.
3. `#5284` — removes the pilot's four back-compat symlinks. Draft until step 1 is deployed.
4. `#5257` … `#5275` — migration layers, bottom-up. After this PR merges, `#5257`'s diff shrinks to just its own layer.
5. `--cleanup` — last, once every job is migrated and a staging replay passed.

Note: `#5257` currently targets `main` (GitHub manages the stack and would not accept a manual base change), so until this PR merges its diff also shows these 14 files.

